PR #61, opened 2016-10-03
docs: api specs edited to match the API spec guide

mchalski:
Note that this is much more than just the layout, below is a changelog of sorts, since the comparison may be difficult even with a diff at hand. 

Layout: 
- besides introducing Michael's guidelines - reworded the verbose api description, moved lots of details to where they IMO belong (param/schema descriptions)

Content:
- removed everything concerning tenant tokens - not supported, not required, could only confuse people
- removed bits about id data encryption - sth for a distant future, currently 100% plain text
- added missing but required 'seq_no' to AuthRequest
- removed the bit about the issued JWT being additionally encrypted with a device's key...IMO it's clearly just a regular token in base64 form
- added missing token standard 'iss' claim
- removed a couple of unimplemented methods

New doc type
- it turned out there are methods which don't belong either to 'integrations' or 'devices'; they're only accessible internally in the Docker network (POST /token/verify, PUT /device/{id}/status)
- they were removed from both docs, a 3rd type of doc was introduced to not lose this info
- side effect - 'integrations' doc is empty now, so I removed it

Observed quirks:
- the returned JWT is 'application/json', but in fact it's a blob of text - could be problematic in the final render

@maciejmrowiec @michaelatmender @bboozzoo @kjaskiewiczz @marekswiecznik


docs/internal_api.yml
```diff
@@ -0,0 +1,133 @@
...
+          description: Internal server error.
+          schema:
+            $ref: '#/definitions/Error'
+  /tokens/verify:
+    post:
+     summary: Check the validity of a token
+     description: |
+        Besides the basic validity check, the checks token expiration time and user-initiated token revocation.
+
+        This method is completely hidden by the API gateway. Services
```
marekswiecznik: partially duplicates info.description

docs/internal_api.yml
```diff
@@ -0,0 +1,133 @@
+swagger: '2.0'
+info:
+  version: '0.1.0'
+  title: Device authentication
+  description: |
+      An API for device authentication handling. Not exposed via the API Gateway - intended for internal use only.
+
+basePath: '/api/0.1.0/'
```
marekswiecznik: missing 'host' attribute, MEN-664

docs/devices_api.yml
```diff
@@ -1,99 +1,121 @@
 swagger: '2.0'
 info:
-  title: Authentication Service Device API
   version: '0.1'
+  title: Device authentication
+  description: |
+    An API for device authentication handling. Intended for use by devices.
 
-host: 'localhost:8080'
```
marekswiecznik: introduced in MEN-664, probably should stay.

docs/devices_api.yml
```diff
@@ -1,99 +1,121 @@
...
+          description: |
+            Authentication successful - a new JWT is issued and returned.
+            The JWT is signed with the API's private key ('RS256' signing algorithm), and contains
+            the following standard claims:
+            * 'iss' - issuer
+            * 'exp' - expiry date
+            * 'sub' - subject (auto-generated device ID)
+            * 'jti' - token's unique identifier (tracked for the purpose of revocation)
+          examples:
+              application/json:  eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.
```
michaelatmender: Since it is returning a blob of text, could this be text/plain, rather than application/json?

michaelatmender:
Layout and style is good! The only thing quirky is application/json example that you mentioned

maciejmrowiec:
> Observed quirks:
> 
> the returned JWT is 'application/json', but in fact it's a blob of text - could be problematic in the final render

Should we open a bug for this?

docs/devices_api.yml
mchalski: what makes sense here - `localhost` or `docker.mender.io`?

mchalski:
@maciejmrowiec yes, here it is:

https://tracker.mender.io/browse/MEN-672

docs/devices_api.yml
bboozzoo: IIRC there's an `application/jwt` for this purpose

docs/devices_api.yml
bboozzoo: I agree with @mchalski , host can be safely removed.

docs/devices_api.yml
```diff
@@ -1,99 +1,121 @@
...
         type: string
         description: The device's public key, generated by the device or pre-provisioned by the vendor.
+      seq_no:
+        type: integer
+        format: uint64
+        description: |
+            An ever-increasing sequence number, recorded by the server
+            upon each authentication request to prevent replay attacks.
+
+            The device may choose any non-negative number for the bootstrap request.
```
bboozzoo: I'm not entirely sure, one could pick 2^64-1 and IIRC this was one of the edge cases. There was a JIRA ticket for discussing this but I can't recall any meaningful conclusions.

docs/internal_api.yml
mchalski: fixed

docs/devices_api.yml
mchalski: fixed

docs/devices_api.yml
mchalski: well, it will quickly break then, that's the only corner case.

anyway, we don't handle it.

should we recommend a small number then? because I don't see a way of solving this, there's always be a possiblity of overflow

docs/internal_api.yml
mchalski: reworded it slightly

docs/devices_api.yml
bboozzoo: @mchalski mentioned that this will be updated when a corresponding change is made to the code. Just for the record I'm ok with that.

docs/devices_api.yml
bboozzoo: Let's leave it at any and see how it plays out.

docs/devices_api.yml
maciejmrowiec: no opinion on that one as long as we are consistent across docs

PR #70, opened 2016-11-22
acceptance tests seqnum

bboozzoo:
@mendersoftware/rndity @GregorioDiStefano 

I've tried to generalize `run-acceptance` so that it could become a part of integration repository rather than that of a specific service. This led to minor changes in `build-acceptance`. The script is supposed to build the image for acceptance tests container. So any image/service specific pieces (like copy of extra files) were moved there.

On top of this, there's a readme detailing running and building processes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -123,13 +123,13 @@ script:
     - sudo docker build -t $DOCKER_REPOSITORY .
 
     # Build acceptance tests container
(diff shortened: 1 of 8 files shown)


bboozzoo:
ping

GregorioDiStefano:
nice work

PR #71, opened 2016-11-29
no key load on request

bboozzoo:
Addresses problems raised in https://tracker.mender.io/browse/MEN-827

### DIFF
--- a/accesslog/middleware.go
+++ b/accesslog/middleware.go
@@ -29,11 +29,12 @@ type AccessLogFormat string
 
 const (
 	DefaultLogFormat = "%t %S\033[0m \033[36;1m%Dμs\033[0m \"%r\" \033[1;30m%u \"%{User-Agent}i\"\033[0m"
-	SimpleLogFormat = "%s %Dμs %r %u %{User-Agent}i"
+	SimpleLogFormat  = "%s %Dμs %r %u %{User-Agent}i"
 )
 
 // AccesLogMiddleware is a customized version of the AccessLogApacheMiddleware.
-// It uses the request-specific custom logger (created by requestlog), which appends the Mender-specific request context.
+// It uses the request-specific custom logger (created by requestlog), which
+// appends the Mender-specific request context.
 type AccessLogMiddleware struct {
 	Format       AccessLogFormat
 	textTemplate *template.Template
--- a/api_devauth.go
+++ b/api_devauth.go
@@ -19,7 +19,6 @@ import (
 	"strings"
 
 	"github.com/ant0ine/go-json-rest/rest"
-	"github.com/mendersoftware/deviceauth/config"
 	"github.com/mendersoftware/deviceauth/log"
 	"github.com/mendersoftware/deviceauth/requestid"
 	"github.com/mendersoftware/deviceauth/requestlog"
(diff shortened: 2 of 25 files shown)


pasinskim:
With regards to the changes you've proposed and implemented it looks like `deviceauth` is the almost exact copy of `useradm` (with the difference of some naming details). Maybe better you guys coordinate beckend-side to do more code sharing (we have a `go-lib-micro` for sharing the smaller packages across backend microservices) instead of copy-pasting. This way we will avoid wasting everyones time to test, debug and fix the same things which are already improved in other microservices. This is not very optimal way of work when we need to repeat the work already done (sometimes without knowing that this was already done elsewhere).

Also, if this is to hard to avoid code being copy-pasted and you are improving the code in one part of the application make sure that the broken code is improved in other parts as well (like have some follow-up task).

The change here was easy to follow, and clearly improvement, but perhaps we should be stricter disallowing the same code being repeated in various places. I can reopen PR and revert the changes so that you can do this properly. Also when you are touching imports (even reordering), perhaps it is a good idea to check for the code that was already moved to `go-lib-micro`. If you are not aware of those changes, perhaps you should talk more to the people who are aware of this so that we are not entering a situation with fixing code which is copy-pasted and fixed somewhere else (again).

You should probably talk and coordinate bit better on how to avoid such a issues.

I will open appropriate tickets for fixes and follow-ups mentioned here.

bboozzoo:
@pasinskim I did not come up with the policy of copy-pasting code so please get your facts straight before adding comments that are clearly incorrect. I have lobbied for a shared library from the very beginning of backend work. The point was brought up again and again. Unfortunately were allowed to have one starting **this October**. Basically the first commit here: https://github.com/mendersoftware/go-lib-micro/commit/d5d3782a7bcfad863ebc9510cecc6c226a10125b so yeah, so much for avoiding copy pasting.

pasinskim:
@bboozzoo I am not claiming you did the copy-paste thing, I've just mentioned that it would be better to avoid doing so and try to reuse instead. The work you did in this PR is just an example of something that was fixed/improved elsewhere, `useradm`, that was left broken here. If we had a shared code, I assume this wouldn't be an issue.

Next, I have not been part of the discussions regarding reusing the code, so I can not say anything here. What is more, if you have already decided to use a shared code (at least in some places) maybe best remove the code which is local to given repository so that we will avoid potential issues (like `log` in this repo).

I've created some tasks for the reference so that we can track this:
https://tracker.mender.io/browse/MEN-860
https://tracker.mender.io/browse/MEN-861

mchalski:
yup, that's basically the story here. 

don't want to play the blame game, but we suggested factoring common stuff out from the very beginning. MaciekM insisted on doing the 'steal&copy' for a good amount of time, we finally came to an agreement on this around october - then `go-lib-micro` came into play. trust me it's clear for us that it's an extremely crappy practice, in fact we were amused it has a brand new, catchy name ;)

as for the core of the issue:
-  we're migrating stuff to `go-lib-micro` whenever we touch new code
-  at this point it has most of the common stuff 
- ...but old services still have to be backported to use it

pasinskim:
@mchalski thanks for clarifying. Perhaps work done in this PR could be an indicator that you've made a good decision.
But as you've mentioned, the decision had been made recently, so maybe you can think of how to make sure that the stuff which is `stolen&copied` from one place and then improved/fixed in one, can be changed in other places so that we are possibly avoiding issues like this one.

PR #76, opened 2016-12-14
Run integration tests using travis

GregorioDiStefano:
Signed-off-by: Greg Di Stefano <greg.distefano+github@gmail.com>


GregorioDiStefano:
@bboozzoo please take a look

GregorioDiStefano:
@maciejmrowiec

.travis.yml
```diff
@@ -237,3 +245,5 @@ deploy:
             repo: $DEPLOY_REPO
             all_branches: true
             condition: $API_DOCS_CHANGED = true
+
```
bboozzoo: empty lines

.travis.yml
```diff
@@ -115,7 +122,8 @@ script:
     # also including their coverage may introduce to much noice. Concentrate on the coverage of local packages.
     # Execute go test on every local subpackage (resolved as dependencies) and generate covreage report for each.
     # Test packages pararell (xargs -P)
-    - go list ./... | grep -v /vendor/ | xargs -n1 -I {} -P 4 go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {}
+
+    - if [ "$JOB_TYPE" = compile_and_basic_tests ]; then go list ./... | grep -v /vendor/ | xargs -n1 -I {} -P 4 go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {}; fi
```
bboozzoo: I think you can actually write:

```
  - |
    if [ "$JOB_TYPE" = compile_and_basic_tests ]; then \
        go list ./... | grep -v /vendor/ | \
            xargs -n1 -I {} -P 4 \
                go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {}; \
    fi
```

and maybe reformat other changed lines this way as well

GregorioDiStefano: I purposely picked a one liner, should i stick to multiline instead?

bboozzoo: Your choice. I'd prefer a multiline version (easier to read without scrolling the already narrow diff view), but others may disagree ;)

kacf: I think I opt for the multiline as well, this is a bit confusing to read.

tests/run-test-environment
```diff
@@ -56,13 +56,14 @@ shift
 
 
 [ "$TEST_TYPE" == "acceptance" ] && {
-    COMPOSE_CMD="docker-compose -p acceptance-tests -f $INTEGRATION_PATH/docker-compose.yml"
+    COMPOSE_CMD="docker-compose -p acceptance-tests -f $INTEGRATION_PATH/docker-compose.yml -f $INTEGRATION_PATH/docker-compose.client.yml -f $INTEGRATION_PATH/docker-compose.demo.yml"
```
bboozzoo: are you sure we need to start client in acceptance tests? IIRC the test emulate a client so we don't need to whole image & vm here

.travis.yml
```diff
@@ -129,7 +130,12 @@ script:
     - git clone https://github.com/mendersoftware/integration.git
 
     # Run dependant microservices and the tests service itself
-    - TESTS_DIR=$PWD/tests ./tests/run-acceptance $PWD/integration ./tests/docker-compose.yml
+    - TESTS_DIR=$PWD/tests ./tests/run-test-environment "acceptance" $PWD/integration ./tests/docker-compose.yml
+
+    # Run dependant microservices and the tests service itself
+    - NO_CLEANUP=true TESTS_DIR=$PWD/tests ./tests/run-test-environment "integration" $PWD/integration ./tests/docker-compose-integration.yml &
```
bboozzoo: don't we want compose up to complete before proceeding to run integration tests?

tests/run-test-environment
kacf: We do, because we also test the real behavior of the client. These tests test the whole chain.

.travis.yml
kacf: The test won't proceed until it has an SSH connection to the client, so it should be safe.

tests/run-test-environment
bboozzoo: Acceptance tests (at least the ones that I wrote and there are no new acceptance tests in this PR) did not require a client running in qemu.

kacf: Oh, sorry, you're right, I thought this was the integration section.

.travis.yml
```diff
@@ -123,13 +132,24 @@ script:
...
+        TESTS_DIR=$PWD/tests ./tests/run-test-environment "acceptance" $PWD/integration ./tests/docker-compose.yml;
+      fi
+
+    # Run dependant microservices and the tests service itself
+    - if [ "$JOB_TYPE" = compile_and_integration_tests ]; then
+        NO_CLEANUP=true TESTS_DIR=$PWD/tests ./tests/run-test-environment "integration" $PWD/integration ./tests/docker-compose-integration.yml &
+      fi
+
+    - if [ "$JOB_TYPE" = compile_and_integration_tests ]; then
+        cd $PWD/integration/tests && bash run.sh;
```
kacf: I'd put parentheses around this to avoid being in the wrong folder later on if we add any steps here:

```
( cd $PWD/integration/tests && bash run.sh )
```

.travis.yml
kacf: Just the multiline stuff remaining, as far as I'm concerned (sorry, I forgot it in the last review).

Except for between the pipes, I don't think you even need the backslashes.

.travis.yml
```diff
@@ -123,13 +132,24 @@ script:
     - sudo docker build -t $DOCKER_REPOSITORY .
 
     # Build acceptance tests container
-    - ./tests/build-acceptance ./tests ./docs/internal_api.yml
+    - if [ "$JOB_TYPE" = compile_and_basic_tests ]; then ./tests/build-acceptance ./tests ./docs/internal_api.yml; fi
```
kacf: And this one.

.travis.yml
kacf: This what I mean:

```
- if [ "$JOB_TYPE" = compile_and_basic_tests ]; then
      go list ./... | \
          grep -v /vendor/ | \
          xargs -n1 -I {} -P 4 go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {}
  fi
```

.travis.yml
kacf: And:

```
- if [ "$JOB_TYPE" = compile_and_basic_tests ]; then
      ./tests/build-acceptance ./tests ./docs/internal_api.yml
  fi
```

GregorioDiStefano:
done

GregorioDiStefano:
@kacf @bboozzoo 

Split integration tests to fast and slow matrix in order to parallelize

GregorioDiStefano:
Something that came up:

Now that we have 3 build jobs (matrix), on merging, they will all push to docker and deploy to S3.

There seems to be no way around this, as mentioned here:
https://github.com/travis-ci/travis-ci/issues/929

But, this isn't that big of a problem, since only PRs that pass will be merged, right?

kacf:
Yes, shouldn't be a big problem since we only upload from the master branch, not from PRs.

PR #80, opened 2017-01-05
do not ship token signing key in container

bboozzoo:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ COPY ./config.yaml /etc/deviceauth/
 
 # example server private key - only for testing purpose
 # this key will be replaced with proper one later
-RUN mkdir /etc/rsa
-COPY ./testdata/private.pem /etc/rsa/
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -5,9 +5,7 @@ COPY ./deviceauth /usr/bin/
 RUN mkdir /etc/deviceauth
 COPY ./config.yaml /etc/deviceauth/
 
-# example server private key - only for testing purpose
-# this key will be replaced with proper one later
+# mount your private key at /etc/deviceauth/rsa/private.pem
```
mchalski: docker-compose also has some commented out volume placeholders - we could add one there, so that's all in one place for reference (writing docs, prod scripts)

bboozzoo: The documentation is part of minio https PR https://github.com/mendersoftware/integration/pull/98 (since I already had this PR opened and had some README changes in there).

PR #84, opened 2017-01-09
relative url in Location header

mchalski:
Issues: MEN-881

using conventions from:
https://tools.ietf.org/html/rfc3986#section-5.4

### DIFF
--- a/api_devauth.go
+++ b/api_devauth.go
@@ -254,11 +254,7 @@ func (d *DevAuthHandler) UpdateDeviceStatusHandler(w rest.ResponseWriter, r *res
 		return
 	}
 
-	devurl := utils.BuildURL(r, uriDevice, map[string]string{
-		":id": devid,
-	})
-
-	w.Header().Add("Location", devurl.String())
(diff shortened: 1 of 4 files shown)


api_devauth.go
```diff
@@ -254,11 +254,7 @@ func (d *DevAuthHandler) UpdateDeviceStatusHandler(w rest.ResponseWriter, r *res
 		return
 	}
 
-	devurl := utils.BuildURL(r, uriDevice, map[string]string{
-		":id": devid,
-	})
-
-	w.Header().Add("Location", devurl.String())
+	w.Header().Add("Location", "../../devices/"+devid)
```
bboozzoo: The endpoint is `/api/0.1.0/devices/<id>/status`, so the URL can be either `../<id>` or `../../devices/<id>`. Not sure if there's any preference here though.

mchalski: right, of course this form can be shortened, and I don't see why not, trying it out atm.

mchalski: done, tests are green

PR #86, opened 2017-01-19
config.yaml: don't override the default 'prod middleware' setting

mchalski:
the dev vs prod middleware is already set up correctly, and controlled
via the 'middleware' setting/'dev' flag (with prod as default). the only problem is the config file
which overrides this setting, and is pulled into the docker image - hence
the stack traces from RecoverMiddleware by default.

Issues: MEN-898

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -10,4 +10,4 @@
 #       production environment
 #
 # Defaults to: prod
-middleware: dev
+#middleware: dev


maciejmrowiec:
maybe this is not the place, but is there a reason not default to `dev` ?

mchalski:
@maciejmrowiec hmm, it can work either way, but that's what you proposed in the jira comment: https://tracker.mender.io/secure/ViewProfile.jspa?name=a10020

mchalski:
to expand - I don't have a strong opinion, but imo this is sound - by default you get a 'clean' version (no rubbish in logs and responses) and have to work extra to get those. the 'extra' work of passing a flag or setting is best transferred to the dev, who knows what they're doing and probably digging into stuff already.

maciejmrowiec:
ack

PR #93, opened 2017-02-01
devauth: support device reauthentication on key & sequence number change

bboozzoo:
Device public key change or sequence number reset is a possible scenario that
may easily be caused by wiping on-device mender client state data.

When this occurs, we will keep rejecting device's authentication request. It is
also not possible to reauthenticate the device, unless user removes it from the
system or it is decomissioned. Unfortunately, none of this is possible right now.

This change allows for device reauthentication if:
- sequence number verification fails
- device public key has changed

If any of these conditions takes place, device status is reset to pending (hence
all authentication will be rejected) and a request is sent to device admission
service. Device has to go through the admission procedure again in order to
get authentication token.

Please discuss.
@estenberg @pasinskim @maciejmrowiec @kacf @mendersoftware/rndity 

Edit: there is not jira ticket that covers this.

Edit2: don't blindly merge it yet

### DIFF
--- a/devauth.go
+++ b/devauth.go
@@ -90,7 +90,6 @@ func (d *DevAuth) SubmitAuthRequestWithClient(r *AuthReq, client requestid.ApiRe
 	id := utils.CreateDevId(r.IdData)
 
 	//check if device exists with the same id+key
-	//TODO at some point add key rotation handling (same id, different key)
 	dev, err := d.findMatchingDevice(id, r.PubKey)
 	if err != nil {
 		d.log.Errorf("find matching device error: %v", err)
@@ -98,12 +97,49 @@ func (d *DevAuth) SubmitAuthRequestWithClient(r *AuthReq, client requestid.ApiRe
 		return "", err
 	}
 
-	//for existing devices - check auth reqs seq_no
+	//for existing devices - check auth reqs seq_no, keys
 	if dev != nil {
-		err = d.verifySeqNo(id, r.SeqNo)
+		seq, err := d.verifySeqNo(id, r.SeqNo)
 		if err != nil {
 			d.log.Errorf("verify seq no error: %v", err)
 			return "", err
+		}
+
+		// reauthentication needed flag
+		var reauth bool
+
+		if !seq {
+			d.log.Infof("detected sequence number mismatch for device %v",
+				dev.Id)
+			// force reauthentication
+			reauth = true
+		}
+
+		if dev.PubKey != r.PubKey {
+			d.log.Infof("detected key change for device %v", dev.Id)
+			// force reauthentication
+			reauth = true
+			// use new public key
+			dev.PubKey = r.PubKey
+
+		}
+
+		if reauth {
+			// reset device status to pending (since it's used later on)
+			dev.Status = DevStatusPending
+
+			// update device in DB, force status to pending
+			if err := d.db.UpdateDevice(&Device{
+				Id:     dev.Id,
+				PubKey: dev.PubKey,
+				Status: DevStatusPending,
+			}); err != nil {
+				return "", errors.Wrap(err, "db update device error")
+			}
+
+			if err := d.cDevAdm.AddDevice(dev, client); err != nil {
+				return "", errors.Wrap(err, "devadm add device error")
+			}
 		}
 	} else {
 		//new device - create in 'pending' state
@@ -172,9 +208,11 @@ func (d *DevAuth) findMatchingDevice(id, key string) (*Device, error) {
 	//both devs nil - new device
 	//both devs !nil - must compare id/key
 	//other combinations: id/key mismatch
-	if devi == nil && devk == nil {
+	switch {
+
+	case devi == nil && devk == nil:
 		return nil, nil
-	} else if devi != nil && devk != nil {
+	case devi != nil && devk != nil:
 		if devi.Id != devk.Id {
 			return nil, ErrDevAuthIdKeyMismatch
 		}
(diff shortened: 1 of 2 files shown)


estenberg:
Would it be possible to Denial of Service with this change?
If I know the ID of a device, can I simply try to authenticate with any public key and the original device will be rejected?

bboozzoo:
Yes, if you forge identity then you will block the original device. Though in that regard, it's quite similar to what you can do with forged identity in the real world :) The new device does not get auto-accepted though, it's in pending state. You have to let it in.

OTOH, by wiping mender client state data you can lock yourself out of the system at the moment. With this, you will be able to recover.

estenberg:
OK, then I would rethink this a bit. We need to assume that the device identities are publicly known, so an unauthenticated DoS attack to disconnect a large fleet of devices would be trivial, unfortunately.

estenberg:
OK, then I would rethink this a bit. We need to assume that the device identities are publicly known, so an unauthenticated DoS attack to disconnect a large fleet of devices would be trivial, unfortunately.

kacf:
Agree, the denial of service is just too easy to carry out in this context, especially since our default one is solely based on the MAC address.

How hard would it be not to decommission the device, but to just add it to the pending queue? I'm thinking a scenario like the following:

1. User discovers that one of his devices is not reporting in. He suspects that some state data has been wiped or corrupted on the device.
2. He removes the key of the device only. IOW all server state data / inventory / etc is still there, but the device has an unknown key.
3. This causes the admission service to list it in the pending queue again, allowing you to accept it as soon as the device calls in with a key.
4. Device resumes reporting.

Dunno if that makes it any easier... Just a suggestion.

bboozzoo:
I'm considering moving all pending requests to a separate namespace. IOW, accepted/rejected devices end up in one collection (i.e. 'known'), all pending devices are in sort of 'staging' collection. Migration from staging to known happens when device is accepted or rejected. 

Frankly, I don't know if it's possible to make this atomic with mongo (i.e. using multiple collections). Maybe we'd need to change DB structure a little.

The problem I'm also facing is that UI for accepting devices is dumbed down. You can't see the device's mac address, nor can you see its key. All you see is device ID (not very helpful) and a timestamp when request was received. So it's hard to figure out what device are you accepting.
The API does provide all necessary bits an pieces though. CLI tools can show the device like this:
```
device: 4a998fb3c750e53375625a9a7e577c9641aac15a1c9cc979817a4eab89998b72
    status: pending
    request time: 2017-02-02T09:19:40.724Z
    attributes:
        mac : 28:d2:44:03:bb:05
    public key:
-----BEGIN PUBLIC KEY-----
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA18F4a3TMwTcuQACxbg7b
e1kULHBv0xV9xyCyVhq7vkjJjP5PnL4rsATZBO7kRmZz1GChls9Tw60pvwzi+pV+
b3EjSkB2t2Nd8//pAG2yRZwbPh08PFxgw3djSfg94kvro0gq2TmPVXRRIzdFWnn8
XVatw4aJvhlaITKr6eA6t0DvwFrtNs+dlZWqtrl4PIi4uhzsJKUU+HQgPNvc/2Ud
sJ2gKUKYA8EkDal/d0QEqiFtP6cwLNl/Epg9bWs3vTWmFo7r+05ItPrlc7oMJN2D
zZ8drInPjx04M8eCoYvolk9B8zD7Cw4pGRsGe/nc8T35TkVwkxoR8+QRx/+uT2ET
hQIDAQAB
-----END PUBLIC KEY-----
    fingerprint: 99:cf:36:4a:f1:48:b7:2a:a5:11:96:81:96:a3:c6:58:0b:79:25:38
```

Anyways, I'm ok with sitting on this a little longer and have it fixed in post beta/1.0.

kacf:
I don't know enough about the database to comment on that, but conceptually I think what you say makes sense.

kjaskiewiczz:
The biggest problem I can see here is that device key is not visible in the UI. Thus, looking at the UI, client cannot differentiate devices with same device identity and different keys.
@eysteins Do we really need to assume that the device identities are publicly known?

kacf:
> Do we really need to assume that the device identities are publicly known?

I believe so. Our current implementation uses MAC address which you can sniff off the network. And even if you use an alternative approach, we can't assume that people will use something which is cryptographically safe. Theoretically we could base the identity on the key, which would make it safe, but then you can never rotate the key, not even in future versions.

kjaskiewiczz:
I thought that this is just for beta/1.0 release, and we are planning to hide this identity data in the future ,e.g. by encrypting data using backend (or device admission) public key.

kacf:
If I understand correctly, whether or not it's encrypted is irrelevant. It's what you use to construct the identity data that matters, and this knowledge, while not exactly public, probably wouldn't be hard to figure out (it would be in the code repository for that device, for example).

But probably best that @estenberg clarifies, yes.

estenberg:
There are a couple of issues being discussed here:

* Giving the user more context for admission (public key in UI). I consider the manual-accept method temporary and for easy smaller scale  - maybe up to a few hundred devices. Users would probably know approximately when devices came online so the timestamp is useful. Even if we added the public key or fingerprint, it is still not very likely that these users would actually verify it. Think through all the work required to do that (log in to all devices, check their public key, compare to what's in the UI). Also see thee task for this here https://tracker.mender.io/browse/MEN-954. The more secure, scalable and long-term solution is to pre-provision the devices in the backend by adjusting or implementing a more sophisticated admission service. So it would be nice to have the (verified) public key fingerprint in the UI in the short term, but I don't think it's critical and see the ticket above for that discussion.

* Secret identity. I think that it is wise to assume that identities will always be publicly known. Like @kacf says, MAC addresses can easily be sniffed. If they use serial numbers, this is probably also quite easy to guess if you know the device and when it was bought. The identity is like your username on Google, for example, you can't break into an account by just knowing that.

* Encrypting identity data. This was proposed as a way for users of our SaaS platform to hide information from devices from *us*. It is still unclear if this is really needed, at least I would consider it more long term and not worry about it for now.

estenberg:
> The problem I'm also facing is that UI for accepting devices is dumbed down. You can't see the device's mac address, nor can you see its key. All you see is device ID (not very helpful) and a timestamp when request was received. So it's hard to figure out what device are you accepting.

This is not true, you can see the device identity data; e.g. the MAC Address before admitting a device today, please see the attachment.
screenshot from 2017-02-02 10-01-04

bboozzoo:
This is what it looked like today in Firefox: http://i.imgur.com/zVzg6mE.png

estenberg:
Uh, that must be a bug - I used Chrome.
Could you bring that up with UX team? :)

bboozzoo:
Let's close this for now and continue the discussion later.

bboozzoo:
I've pulled the latest images, the UI is fixed now.

PR #96, opened 2017-02-06
api: internal: no See Other on device status PUT

bboozzoo:
Do not return 303 See Other on PUT to /api/0.1.0/devices/:id/status as this
would redirect to an endpoint that no longer exists.

### DIFF
--- a/api_devauth.go
+++ b/api_devauth.go
@@ -235,8 +235,7 @@ func (d *DevAuthHandler) UpdateDeviceStatusHandler(w rest.ResponseWriter, r *res
 		return
 	}
 
-	w.Header().Add("Location", "../"+devid)
-	w.WriteHeader(http.StatusSeeOther)
+	w.WriteHeader(http.StatusNoContent)
 }
 
 // return selected http code + error message directly taken from error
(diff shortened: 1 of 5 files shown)


api_devauth.go
```diff
@@ -235,8 +235,7 @@ func (d *DevAuthHandler) UpdateDeviceStatusHandler(w rest.ResponseWriter, r *res
 		return
 	}
 
-	w.Header().Add("Location", "../"+devid)
-	w.WriteHeader(http.StatusSeeOther)
+	w.WriteHeader(http.StatusOK)
```
maciejmrowiec: on other API's we use 204 `No content` for successful PUT requests

bboozzoo: Yeah, maybe. Changing to 204 would require a chage in deviceadm too.

maciejmrowiec: Won't be able to change this later, once released will stay with us for long time. It's nice to keep it consistent, makes it more polished.

maciejmrowiec: Btw nice catch with this issue, quite tricky to spot.

bboozzoo: updated to 204

bboozzoo:
ok, i've opened a PR in deviceadm too, both builds will fail now. We'll have to merge one (possibly deviceadm first) and then this one.

kjaskiewiczz:
I'm not sure why endpoint /api/0.1.0/devices/:id is not implemented.
IMO we should implement this endpoint and use it at least in the UI

Current solution looks like workaround for me.

maciejmrowiec:
yep, definitely we need to get those at some point, once we have it we can even bring back the location header

PR #111, opened 2017-03-03
identity split

bboozzoo:

Opening PR early. Acceptance tests are likely to fail. Quite possible that integration tests will fail too.


bboozzoo:
Ok, most of the changes is there. I think we might want to go over them offline as there's quite a bit that got updated.

The core of the changes is summarized below:
* introduced `AuthSet` - a wrapper for authentication data set
** auth set is composed of device key and identity, this set will be verified by the user and accepted/rejected, each set is acted on separately
** a combination of device identity & key must be unique within authset collection
* device
** device identity must be unique across devices collection, we cannot have more than one device with particular identity, identity is supposed to be unique (i.e. composed of such and so many attributes that it becomes unique)
** status field lost its meaning, when giving out tokens devauth will look on the status of auth set (this makes sense as auth set is identity + key that was verified by the user, and we give out tokens if device presents this particular set of attributes)
* tokens are tied to auth sets, although only through DB since `sub` field is used for device ID in other services, so I could not reuse that field and change its meaning

Now the bad part:
* added `AuthSet` in management API definitions
* modified `Device` API definition to include `AuthSet` (both `/devices` and `/devices/{id}` include this)
* TBD but does this warrant internal API bump?
* `/devices/{id}/status` - {id} has special meaning, and means authentication data set ID, while one would expect it to mean device ID (just like for `/devices/{id}`)
** this is to preserve backwards compatibility with deviceadm, but IMO we should break the API right now to avoid hurdles later on (there's no known users, and only 1 potential user who we can assist with upgrading their code)
** ideally we should have `/devices/{id}/auth/{auth_id}/status` endpoint which deviceadm will PUT to
* found sort-of device provisioning task, now marked with TODO in the code, but should be moved to orchestrator once we have one

*NO* migration support at this point.

review please

bboozzoo:
Currrently blocked by https://github.com/mendersoftware/deviceadm/pull/75 https://github.com/mendersoftware/deviceadm/pull/74

datastore_mongo.go
```diff
@@ -290,6 +290,95 @@ func (db *DataStoreMongo) Index() error {
 		return err
 	}
 
+	// auth requests
+	err = s.DB(DbName).C(DbAuthSetColl).EnsureIndex(mgo.Index{
+		Unique: true,
+		// identity data shall be unique within collection
```
bboozzoo: done, fixup & rebase

devauth.go
```diff
@@ -84,36 +84,69 @@ func (d *DevAuth) SubmitAuthRequest(r *AuthReq) (string, error) {
...
+	if err != nil {
+		d.log.Error("failed to find device auth set but could not add one either")
+		return "", errors.New("failed to locate device auth set")
+	}
 
-		if err := d.cDevAdm.AddDevice(dev, client); err != nil {
+	// it it was indeed added (a new request), pass it to admission service
+	if added || !to.Bool(areq.AdmissionNotified) {
+		if err := d.cDevAdm.AddDevice(dev, areq, client); err != nil {
+			// we've failed to submit the request, dump it from db
```
bboozzoo: done, fixup & rebase

bboozzoo:
the tests are green, merge?

PR #128, opened 2017-04-27
use per tenant database in the mongo datastore

kjaskiewiczz:
Issues: MEN-1051


store/mongo/datastore_mongo.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/mendersoftware/deviceauth/store"
 
 	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
+	ctxStore "github.com/mendersoftware/go-lib-micro/store"
```
bboozzoo: can you rename this to ctxstore?

kjaskiewiczz: According to https://golang.org/doc/effective_go.html#names
both names are not very good, because 
"there should be no need for underscores or mixedCaps".
ctxstore is not single-word name so there is a need for mixedCaps IMO.

bboozzoo: Hmm, ok if you feel that `ctxStore` is ok, then feel free to merge it.

PR #130, opened 2017-05-12
multi tenant token verifaction & device authentication

bboozzoo:

 There's still some work to be done, but I'm opening the PR for review anyway.


bboozzoo:
Looks like this is fairly complete. I'm still switching to `go-lib-micro/apiclient`, so expect another rebase or two.

bboozzoo:
review anyone?

mchalski:
@bboozzoo see https://github.com/mendersoftware/deviceauth/pull/132#issuecomment-302399256

at least one client gets a `requestid.ApiRequester` (in form of `http.Client`)  instead of `apiclient.HttpRunner` (or, in fact, its implementation `HttpApi`). the whole thing seems to 'work' because these interfaces are actually the same, but the result is very different - the headers are not propagated.

PR #131, opened 2017-05-15
decommissioning fix

bboozzoo:
Details are covered in commit 
https://github.com/mendersoftware/deviceauth/commit/41d4aec24a25521719661a0b1a20e4de09d694aa

Note that there is still a chance that device will be created in inventory if attributes PATCH request goes through token verification before decommissioning commenced but the actual requested is proxied to inventory after DELETE device was run. Fixing this would require a change in how PATCH on device attributes works.

@mendersoftware/rndity @maciejmrowiec @GregorioDiStefano 

Best if verifed with https://github.com/mendersoftware/mender-api-gateway-docker/pull/64 otherwise correlating all logs with their corresponding requests may not be accurate.

### DIFF
--- a/devauth/devauth.go
+++ b/devauth/devauth.go
@@ -255,8 +255,10 @@ func (d *DevAuth) DecommissionDevice(ctx context.Context, devId string) error {
 	l.Warnf("Decommission device with id: %s", devId)
 
 	// set decommissioning flag on the device
-	updev := &model.Device{Id: devId, Decommissioning: true}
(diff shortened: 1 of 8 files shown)


bboozzoo:
merge?

maciejmrowiec:
should we merge? we are about to branch 1.1

GregorioDiStefano:
yes, decommissioning doesn't work well without this fix.

PR #132, opened 2017-05-17
[DON'T MERGE] propagate auth header to deviceadm on PUT /devices/:id

mchalski:
Issues: MEN-1068

the change is rather small, and mostly required careful tracking of all our context-juggling magic in middlewares and api clients. summary:
-  introduce and use a generic api client HttpRunner/HttpApi, which passes both request id and authorization header into outgoing service requests
- the runner etc. was factored out into `go-lib-micro/apiclient`, PR: https://github.com/mendersoftware/go-lib-micro/pull/24, merge this first
- pack the authorization header into request context via `go-lib-micro/context` middleware
- cleanup spurious stuff

this pattern of passing auth data to other services is consistent with @bboozzoo 's recent `deviceauth` PR (https://github.com/mendersoftware/deviceauth/pull/130), and with code already present in `deviceadm`. we should aim at making this consistent across the board.


mchalski:
@bboozzoo almost : https://github.com/bboozzoo/mender-deviceauth/blob/0f33dbf10f446db24a79ef5421a8e4307ac85026/devauth/devauth.go#L51

simpleApiClientGetter still returns a vanilla `http.Client`; to make the headers work it should return `apiclient.HttpApi` (also we could get rid of requestid.ApiRequester).

bboozzoo:
Uhh, right, missed that. I suggest to merge #130 and rebase your PR on top of that.

bboozzoo:
I'm taking over this PR, will rebase and open a new one.

PR #146, opened 2017-08-25
reject request with tenant token in single tenant setup

kjaskiewiczz:
https://tracker.mender.io/browse/MEN-1327


kjaskiewiczz:
After implementing this, I'm not sure if we really want to reject auth requests with tenant token in single tenant scenario.
Tenant token could be something that the device is sending all the time, but is interpreted only by multi tenant setup, maybe?

PR #153, opened 2017-09-20
limiting maximum number of accepted devices per tenant - internal end…

PiotrPrzybylak:
…point spec

Issues: MEN-1419

Changelog: None

Signed-off-by: Piotr Przybylak <piotr.przybylak@rndity.com>

### DIFF
--- a/docs/internal_api.yml
+++ b/docs/internal_api.yml
@@ -38,7 +38,77 @@ paths:
             description: Unexpected error.
             schema:
               $ref: '#/definitions/Error'
+  /tenant/{tenant_id}/limits/max-device-count:
+    get:
+      summary: Max device count limit
+      parameters:
+        - name: tenant_id
+          in: path
+          type: string
+          description: Tenant ID.
+          required: true
+      responses:
+        200:
+          description: Successful response.
+          schema:
+            title: MaxDeviceCountLimit
+            schema:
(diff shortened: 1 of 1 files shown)


docs/internal_api.yml
```diff
@@ -38,7 +38,76 @@ paths:
             description: Unexpected error.
             schema:
               $ref: '#/definitions/Error'
+ /tenant/{tenant_id}/limits/max_device_count:
+    get:  Max device count limit
+      summary: List known tenants
```
bboozzoo: summary is incorrect

docs/internal_api.yml
```diff
@@ -38,7 +38,76 @@ paths:
             description: Unexpected error.
             schema:
               $ref: '#/definitions/Error'
+ /tenant/{tenant_id}/limits/max_device_count:
```
bboozzoo: can we replace `_` with `-` ?

bboozzoo: Do we need `/tenant/{tenant_id}/limits` as well? (would return all currently set limits)

bboozzoo: Also, I don't think there's any other internal endpoint that takes `tenant ID` explicitly. I suppose they all take `Authorization` header with a proper tenant token (i.e. token with `mender.tenant` claim).

PR #156, opened 2017-09-22
migration fix

bboozzoo:
@mendersoftware/rndity comments & review please

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/mendersoftware/go-lib-micro/identity"
 	"github.com/mendersoftware/go-lib-micro/log"
 	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
 	ctxstore "github.com/mendersoftware/go-lib-micro/store"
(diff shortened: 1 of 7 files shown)


store/mongo/datastore_mongo.go
```diff
@@ -350,7 +351,25 @@ func (db *DataStoreMongo) Migrate(ctx context.Context, version string) error {
 	}
 
 	for _, d := range dbs {
-		l.Infof("migrating %s", d)
+		// if not in multi tenant, then tenant will be "" and identity
```
mchalski: I'd avoid more context magic like the plague.

we're in a loop over valid tenant dbs, we have their names, migrations could just accept a verbatim dbname.

we're doing extra work by shuffling the name into a context, so that the migration can unpack it.

I won't request changes, but don't count on me doing this in any new service ;)

bboozzoo:
Uhh, integration failed with: 
```
(*AccessLogMiddleware).MiddlewareFunc.func1" line=60 request_id=fa1143ee-2d5f-4628-b6b1-0d77c3fa91ba 
An HTTP request took too long to complete. Retry with --verbose to obtain debug information.
If you encounter this issue regularly because of slow network conditions, consider setting COMPOSE_HTTP_TIMEOUT to a higher value (current value: 300).
```
Any ideas what to do now? @GregorioDiStefano ?

bboozzoo:
merge?

PR #157, opened 2017-09-22
get max devices limit (mgmt endpoint)

mchalski:
issues: https://tracker.mender.io/browse/MEN-1423

the first 3 commits actually unify the name of the `max_devices` param across the codebase.

### DIFF
--- a/devauth/devauth.go
+++ b/devauth/devauth.go
@@ -72,6 +72,8 @@ type App interface {
 	VerifyToken(ctx context.Context, token string) error
 
 	SetTenantLimit(ctx context.Context, tenant_id string, limit model.Limit) error
+
+	GetLimit(ctx context.Context, name string) (*model.Limit, error)
 }
 
 type DevAuth struct {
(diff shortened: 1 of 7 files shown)


devauth/devauth.go
```diff
@@ -573,6 +575,19 @@ func (d *DevAuth) VerifyToken(ctx context.Context, raw string) error {
...
 }
 
+func (d *DevAuth) GetLimit(ctx context.Context, name string) (*model.Limit, error) {
+	lim, err := d.db.GetLimit(ctx, name)
+
+	switch err {
+	case nil:
+		return lim, nil
+	case store.ErrLimitNotFound:
+		return &model.Limit{Name: name, Value: 0}, nil
```
bboozzoo: `nil, nil`?

mchalski: well, if the limit is not there, we set it to 0, no?

and later on, to the config default (spearate task)

bboozzoo: fair enough

PR #161, opened 2017-10-02
GET /devices/count?status

mchalski:
https://tracker.mender.io/browse/MEN-1484

involves a pretty horrific mongo aggregation pipeline (to count devices by status), so take a good look.
can filter by status in general, not only 'accepted' devices. this required more work but thought it would be useful to handle this once and for all and not go back to it. other statuses might still be relevant e.g. for the ui in the future.

also, had to bump mongo version on travis to 3.4. we're still stuck at 3.2 (which imo should be fixed across the board). I got bitten by an incompatibility - the `$count` aggregate operator is available since 3.4, hence the fix.

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -605,3 +605,125 @@ func (db *DataStoreMongo) GetLimit(ctx context.Context, name string) (*model.Lim
 
 	return &lim, nil
 }
+
+func (db *DataStoreMongo) GetDevCountByStatus(ctx context.Context, status string) (int, error) {
+	s := db.session.Copy()
+	defer s.Close()
+
+	// if status == "", fallback to a simple count of all devices
+	if status == "" {
+		c := s.DB(ctxstore.DbFromContext(ctx, DbName)).C(DbDevicesColl)
+		return c.Count()
+	}
+
+	// compose aggregation pipeline
+	c := s.DB(ctxstore.DbFromContext(ctx, DbName)).C(DbAuthSetColl)
+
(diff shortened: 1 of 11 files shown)


mchalski:
@mendersoftware/rndity ping?

bboozzoo:
rebase?

store/mongo/datastore_mongo.go
```diff
@@ -595,3 +595,125 @@ func (db *DataStoreMongo) GetLimit(ctx context.Context, name string) (*model.Lim
 
 	return &lim, nil
 }
+
+func (db *DataStoreMongo) GetDevCountByStatus(ctx context.Context, status string) (int, error) {
```
maciejmrowiec: this looks a bit complicated, maybe I'm not getting full intention but that was supposed to count devices of specified status, devices already have status field so there is no need to mine authsets (is there?)

then something like this ```db.devices.find({"status":"accepted"}).count()``` could do the trick

mchalski: the thing is the `device.status` field is obsolete and not consistently updated since authsets were introduced. the idea I got from talking to the rest of the team is that we will drop it at some point.

so auth sets are the actual 'source of truth' on the overall status.

the query is complicated because I did too much - I didn't notice that only `accepted` devices must be counted for now, and automatically assumed we're after all statuses. it was pointed out to me later, and since I already started composing the query I didn't want to drop it - I imagine at some point counting other statuses will be useful.

maciejmrowiec: I see, thanks for explanation.

PR #169, opened 2017-10-09
tests: add test for limiting of accepted devices 

bboozzoo:
@mendersoftware/rndity

### DIFF
--- a/tests/tests/test_device.py
+++ b/tests/tests/test_device.py
@@ -120,6 +120,26 @@ class TestDevice:
         print('limit:', limit)
         assert limit.limit == 0
 
+    @pytest.mark.xfail(reason='Not implemented yed, waiting for MEN-1486')
+    @pytest.mark.parametrize('devices', ['5'], indirect=True)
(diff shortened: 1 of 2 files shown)


tests/tests/test_device.py
```diff
@@ -120,6 +120,27 @@ def test_get_device_limit(self, management_api):
...
+    @pytest.mark.xfail(reason='Not implemented yed, waiting for MEN-1486')
+    @pytest.mark.parametrize('devices', ['5'], indirect=True)
+    def test_device_limit_applied(self, management_api, devices):
+        expected = 2
+        internal_api.put_max_devices_limit('foo', expected)
+
+        accepted = 0
+        try:
+            for dev, dev_auth in devices:
+                with inventory.run_fake_for_device_id(devid):
```
kjaskiewiczz: where this variable (devid) came from?

bboozzoo: Right, this probably failed due to NameError ;) The test was expected to fail, just not the way it actually failed

bboozzoo: should be fixed now

PR #170, opened 2017-10-10
enforce device count limit

mchalski:
https://tracker.mender.io/browse/MEN-1486

this is part 1 of the work. next PR will go into `deviceadm` (propagate the new 422 status to the user when accepting a device, with the exact same message).

### DIFF
--- a/devauth/devauth.go
+++ b/devauth/devauth.go
@@ -39,8 +39,9 @@ import (
 )
 
 var (
-	ErrDevAuthUnauthorized = errors.New("dev auth: unauthorized")
-	ErrDevIdAuthIdMismatch = errors.New("dev auth: dev ID and auth ID mismatch")
+	ErrDevAuthUnauthorized   = errors.New("dev auth: unauthorized")
+	ErrDevIdAuthIdMismatch   = errors.New("dev auth: dev ID and auth ID mismatch")
+	ErrMaxDeviceCountReached = errors.New("maximum number of accepted devices reached")
 )
 
 // Expiration Timeout should be moved to database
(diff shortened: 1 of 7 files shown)


bboozzoo:
can you rebase on top of master and enable this test https://github.com/mendersoftware/deviceauth/blob/master/tests/tests/test_device.py#L125 ?

mchalski:
@mendersoftware/rndity this finally builds, ping

devauth/devauth.go
```diff
@@ -403,6 +404,15 @@ func (d *DevAuth) DecommissionDevice(ctx context.Context, devId string) error {
 }
 
 func (d *DevAuth) AcceptDeviceAuth(ctx context.Context, device_id string, auth_id string) error {
+	allow, err := d.canAcceptDevice(ctx)
```
bboozzoo: Here's the thing, in order to correctly apply the limit, we should first 'accept' the device and then verify if it's above the limit. Otherwise there's a race between checking agains the and updating a shared state in DB. IMO it will be possible to abuse it if you spam the status endpoint.

That's the theory. In practice, we've cut some corners already so this one might be acceptable too (even though it's wrong). Maybe at least leave a note in the code that this part is racy and we'll need to do something about this in the future.

mchalski: so, should the flow be like this?:
- mark auth set as accepted
- count
- if above limit - unmark?

that's 'better', but nontransactionality could still bite us. not sure if it's worth the effort, but I'll consider it.

mchalski: @bboozzoo added a note ;)

PR #174, opened 2017-10-23
correctly set content-type of JWT auth token

GregorioDiStefano:
Noticed that the token response headers are incorrect this weekend while investigating something else.
The response content-type is "application/json" but the response is not valid JSON  

@bboozzoo @maciejmrowiec

### DIFF
--- a/api/http/api_devauth.go
+++ b/api/http/api_devauth.go
@@ -155,8 +155,9 @@ func (d *DevAuthApiHandlers) SubmitAuthRequestHandler(w rest.ResponseWriter, r *
 			http.StatusUnauthorized, "unauthorized")
 		return
 	case nil:
-		w.(http.ResponseWriter).Header().Set("Content-Type", "application/jwt")
-		w.(http.ResponseWriter).Write([]byte(token))
+		ww := w.(http.ResponseWriter)
+		ww.Header().Set("Content-Type", "application/jwt")
+		ww.Write([]byte(token))
 		return
 	default:
 		rest_utils.RestErrWithLogInternal(w, r, l, err)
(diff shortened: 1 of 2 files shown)


GregorioDiStefano:
that didn't fix it, taking another look

bboozzoo:
Might be that go-json-rest is forcing `application/json` content type on the output

maciejmrowiec:
Yeah, I think auto-setting application/json is in middleware

GregorioDiStefano:
Yes, that's whats happening! 
A quick suggestion: if I add a quote around the token, it would be valid JSON. 

Is that an acceptable change?

bboozzoo:
Although it should not happen: https://github.com/ant0ine/go-json-rest/blob/master/rest/response.go#L67

GregorioDiStefano:
The unit tests are actually testing the content-type, and are passing "application/jwt" is correctly being set, but the acceptance tests in python are seeing the content-type as "application/json; charset=utf-8"

bboozzoo:
There's a unit test that verifies if content type is `application/jwt`: https://github.com/mendersoftware/deviceauth/blob/master/api/http/api_devauth_test.go#L261 idk, maybe recorder is hooked up in the wrong place.

Also, keep in mind that errors will return with content-type set to `application/json`

GregorioDiStefano:
If I understood correctly, this request is "internal" and actually handled in useradm:

see here:
https://github.com/mendersoftware/useradm/blob/master/api/http/api_useradm.go#L115

GregorioDiStefano:
Fixed here: https://github.com/mendersoftware/useradm/pull/100

PR #178, opened 2017-11-27
docs/internal: preauth endpoint definition

mchalski:
changelog: none


docs/internal_api.yml
```diff
@@ -108,6 +108,38 @@ paths:
           description: Internal server error.
           schema:
            $ref: "#/definitions/Error"
+  /devices:
+    post:
+      summary: Submit a preauthorized device.
+      description: |
+          Adds a given device/authentication data set in the 'preauthorized' state.
+          Designed to be called from admission, with precomputed device_id and auth_set_id.
```
maciejmrowiec: What do you think about instead taking precomputed id's, return id's back to caller? This way we could get single place that is responsible for ID assigning, or it's not worth it?

mchalski: not worth it imo. we create ids by simple `mgo.NewObjectId.Hex()`.

PR #179, opened 2017-11-27
docs: management endpoint for removing device authentication sets

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-1633

### DIFF
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -110,6 +110,33 @@ paths:
           description: Internal server error.
           schema:
             $ref: "#/definitions/Error"
+  /devices/{id}/auth/{aid}:
+    delete:
+      summary: Remove the device authentication set
+      parameters:
+        - name: Authorization
+          in: header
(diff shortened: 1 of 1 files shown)


docs/management_api.yml
```diff
@@ -110,6 +110,33 @@ paths:
...
+          description: Device identifier.
+          required: true
+          type: string
+        - name: aid
+          in: path
+          description: Authentication data set identifier.
+          required: true
+          type: string
+      responses:
+        204:
```
mchalski: if it's management, are we considering 401 here? or is it 'internal management', meaning it will only be called via deviceadm, so auth will lready be verified?

kjaskiewiczz: It's management. We are not including 401 in responses of any of the endpoints.
I think that we have information in the docs that every management endpoint can return 401.

mchalski: oh ok.

PR #190, opened 2018-01-12
devauth: fix preauth case handling in POST /auth_requests 

mchalski:
set the 'accepted' status just before returning to the main routine -
but not before, the struct is being used to match a 'preauthorized' entry
in the db ('accepted' will never match).

changelog: none


mchalski:
integration: https://mender-jenkins.mender.io/job/mender-builder/140/console

PR #197, opened 2018-02-15
devauth: fix auth set removal

mchalski:
don't remove the device because that might leave dangling data
in other services with no possibility of decommissioning/removal.

only delete the device if the authset was preauthorized - that's a guarantee
data was not propagated yet anywhere else.

changelog: none

### DIFF
--- a/devauth/devauth.go
+++ b/devauth/devauth.go
@@ -484,8 +484,7 @@ func (d *DevAuth) DecommissionDevice(ctx context.Context, devId string) error {
 	return d.db.DeleteDevice(ctx, devId)
 }
 
-// Deletes device authentication set.
-// If there is only one authentication set for the device, the device will also be deleted.
+// Deletes device authentication set, and optionally the device.
 func (d *DevAuth) DeleteAuthSet(ctx context.Context, devId string, authId string) error {
 
 	l := log.FromContext(ctx)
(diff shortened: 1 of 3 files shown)


devauth/devauth.go
```diff
@@ -484,8 +484,7 @@ func (d *DevAuth) DecommissionDevice(ctx context.Context, devId string) error {
 	return d.db.DeleteDevice(ctx, devId)
 }
 
-// Deletes device authentication set.
-// If there is only one authentication set for the device, the device will also be deleted.
+// Deletes device authentication set, and otionally the device.
```
kjaskiewiczz: s/otionally/optionally

tests/tests/test_device.py
```diff
@@ -298,6 +298,38 @@ def _test_delete_authset_OK(self, management_api, devices, **kwargs):
...
+        assert dev.auth_sets[0].id == aid
+
+        # delete auth set, check device deleted
+        rsp = management_api.delete_authset(dev.id, aid, **kwargs)
+        assert rsp.status_code == 204
+
+        try:
+            found = management_api.get_device(id=device_id, **kwargs)
+        except bravado.exception.HTTPError as e:
+            assert e.response.status_code == 404
```
kjaskiewiczz: no error == no assertion

tests/tests/test_device.py
```diff
@@ -298,6 +298,38 @@ def _test_delete_authset_OK(self, management_api, devices, **kwargs):
...
+
+        # check all other devices intact
+        after_devs = management_api.list_devices(**kwargs)
+        assert len(after_devs) == len(devices)
```
kjaskiewiczz: it only checks if one of the devices was deleted (if there will be no 404 error on delete and different device has been deleted, test will pass)

PR #203, opened 2018-03-21
Remove tenant/device tokens

mchalski:
Issues: https://tracker.mender.io/browse/MEN-1696


docs/internal_api.yml
```diff
@@ -40,12 +40,12 @@ paths:
               $ref: '#/definitions/Error'
   /tokens:
     delete:
-      summary: Delete device tokens. for a tenant with given tenant id
+      summary: Delete device tokens for a tenant with given tenant_id
```
kjaskiewiczz: Just "Delete device tokens" maybe?
Whole sentence is not exactly true since we added optional device_id parameter.

mchalski: fixed

PR #214, opened 2018-07-02
migrate PUT /admission/devices/:id/status

mchalski:
https://tracker.mender.io/browse/MEN-1923

### DIFF
--- a/docs/management_api_admission.yml
+++ b/docs/management_api_admission.yml
@@ -1,0 +1,99 @@
+swagger: '2.0'
+
+info:
+  version: '1'
+  title: Device admission
+  description: |
+    An API for device admission handling. Intended for use by the web GUI.
+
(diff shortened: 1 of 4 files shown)


docs/management_api_admission.yml
```diff
@@ -0,0 +1,95 @@
...
+              status: "accepted"
+        400:
+          description: |
+              The request body is malformed or the state transition is invalid. See error for details.
+          schema:
+            $ref: "#/definitions/Error"
+        404:
+          description: The device authentication data set was not found.
+          schema:
+            $ref: "#/definitions/Error"
```
kjaskiewiczz: it is possible that this endpoint will return 422 Unprocessable Entity (when user will hit max device count limit)
this should also be documented

mchalski: fixed

PR #216, opened 2018-07-06
GET /admission/devices

mchalski:
https://tracker.mender.io/browse/MEN-1926

### DIFF
--- a/api/http/api_devauth.go
+++ b/api/http/api_devauth.go
@@ -50,6 +50,7 @@ const (
 
 	// migrated devadm api
 	uriDevadmAuthSetStatus = "/api/management/v1/admission/devices/:aid/status"
+	uriDevadmDevices       = "/api/management/v1/admission/devices"
 
 	HdrAuthReqSign = "X-MEN-Signature"
 )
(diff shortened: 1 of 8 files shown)


api/http/api_devauth_test.go
```diff
@@ -1720,3 +1721,157 @@ func TestApiDevAuthGetTenantDeviceStatus(t *testing.T) {
...
+	for _, v := range rec.Header()[hdr] {
+		if v == val {
+			return v
+		}
+	}
+
+	return ""
+}
+
+func ToJson(data interface{}) string {
```
kjaskiewiczz: maybe ToJsonString instead of ToJson? since it's returning JSON string

api/http/api_devauth.go
```diff
@@ -50,13 +50,16 @@ const (
 
 	// migrated devadm api
 	uriDevadmAuthSetStatus = "/api/management/v1/admission/devices/:aid/status"
+	uriDevadmGetDevices    = "/api/management/v1/admission/devices"
```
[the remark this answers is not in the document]
mchalski: fixed

api/http/api_devauth_test.go
mchalski: fixed

PR #229, opened 2018-09-04
docs: introduce version 2 of the management API 

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-1981

Please take a look really carefully.
There is no GET /devices/{id}/auth/{aid} endpoint in the new API since I can't see a point in having this.


docs/management_api_v2.yml
```diff
@@ -0,0 +1,507 @@
...
+
+basePath: '/api/management/v2/devauth/'
+host: 'mender-device-auth:8080'
+schemes:
+  - http
+
+paths:
+  /devices:
+    get:
+      summary: Get a list of tenant's devices.
```
michaelclelland: Is the list sorted? It might be useful to sort by created_ts or updated_ts

docs/management_api_v2.yml
```diff
@@ -0,0 +1,507 @@
...
+      description: |
+        Provides a list of tenant's devices, with optional device status filter.
```
michaelatmender: Is this list sorted? It might be useful to sort by created_ts or updated_ts

kjaskiewiczz: currently, in the deviceauth and deviceadm services, we are sorting devices by id
to me, it sounds reasonable to sort them by created_ts
@mchalski @maciejmrowiec what do you think?

maciejmrowiec: 👍

docs/management_api_v2.yml
```diff
@@ -0,0 +1,507 @@
...
+    type: object
+    properties:
+      error:
+        description: Description of the error
+        type: string
+  PreAuthSet:
+    type: object
+    properties:
+      device_identity:
+        type: string
```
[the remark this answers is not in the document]
kjaskiewiczz: good idea

kjaskiewiczz:
Comments addressed by adding two new commits.

PR #239, opened 2018-10-09
docs: management API v2: POST devices/: return conflicting device

kjaskiewiczz:
When there is a conflict while preauthorizing devices, return
conflicting device in 409 Conflict response body.


maciejmrowiec:
looks good

PR #251, opened 2018-10-29
store/mongo: fix migration 1.4.0

kjaskiewiczz:
Migration 1.4.0 is setting device statuses based on devices
authentication set statuses.
If the device has no authentication sets then set device status to
"rejected" (there is no better status in this case).

### DIFF
--- a/store/mongo/migration_1_4_0.go
+++ b/store/mongo/migration_1_4_0.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/mendersoftware/deviceauth/model"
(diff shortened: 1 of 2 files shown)


maciejmrowiec:
@kjaskiewiczz this one passed integration, but travis unit/acceptance need some fixing

There is change-log missing from commit (I recommend to add it to commit message template in git configuration if you don't have it yet)

maciejmrowiec:
Reopened this PR with added changelog in the message and rebase, closing.
NEW PR: https://github.com/mendersoftware/deviceauth/pull/253

PR #253, opened 2018-11-06
store/mongo: fix migration 1.4.0

maciejmrowiec:
Migration 1.4.0 is setting device statuses based on devices
authentication set statuses.
If the device has no authentication sets then set device status to
"rejected" (there is no better status in this case).

changelog: none

Signed-off-by: Krzysztof Jaskiewicz <krzysiekjaskiewicz@gmail.com>


maciejmrowiec:
blocked by timeout only

maciejmrowiec:
Travis build passed but failed to report. 
Integration looks good. Merging.

PR #254, opened 2018-11-07
Move /devices/count routing rule before /devices/:id to match "count"…

maciejmrowiec:
… properly.

Definition order matters as it's a fall though mechanism. Use first matching one, in this case 'count' was considered as :id

Changelog: none

Signed-off-by: Maciej Mrowiec <mrowiec.maciej@gmail.com>


maciejmrowiec:
@kjaskiewiczz  just FYI quick catch

PR #261, opened 2018-11-27
changelog: docs: API v1 deprecation note

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-1987

### DIFF
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -4,6 +4,7 @@ info:
   title: Device authentication
   description: |
       An API for device authentication handling.
+      This API version has been deprecated and will be removed in Mender 1.8 release.
 
 basePath: '/api/management/v1/devauth/'
 host: 'mender-device-auth:8080'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ language: go
 # Forces travis to use VM insted container, required to be able to build containers.
 sudo: required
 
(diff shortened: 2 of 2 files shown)


docs/management_api.yml
```diff
@@ -4,6 +4,7 @@ info:
   title: Device authentication
   description: |
       An API for device authentication handling.
+      This API version has been deprecated and will be removed in Mender 1.8 release.
```
maciejmrowiec: Admission needs to get also a note

kjaskiewiczz: yes; I need to add a note to 1.6 admission api and remove admission api from master/1.7

kjaskiewiczz: 1.6 release contains admission 1.4, so is it ok to just add a note to admission 1.4? or do I need to create separate deviceadm branch - 1.6.x, add a note there and update integration 1.6 to contain admission 1.6?

kacf:
I'm not sure why the license test started failing now, that requirement has been there for quite a while already. But the solution is easy, just update the copyright year in all the files it failed on.

docs/management_api.yml
kacf: I think you mean [this one](https://github.com/mendersoftware/deviceadm/commit/54185dae47189d7db3c4bdd07bce8cb6f391de5a), right? I submitted it yesterday.

PR #266, opened 2018-11-30
api/http: log warnings about deprecated endpoints

kjaskiewiczz:


kjaskiewiczz:
integration is failing but not because of this change
there is an issue with conductor a python worker;
thus I'm planning to merge this PR without waiting for integration to start working

PR #278, opened 2019-03-05
changed mapreduce to pipe

briannosaurus:
Changed this to allow the use of Mongo Atlas as they do not allow mapreduce on their lower tiers.


briannosaurus:
can you close this pull request and i will make a new one with the proper signoff and git-changelog showing i tested heavily before committing.

PR #284, opened 2019-03-07
Changed mapreduce to pipe so it can be used with Mongo Atlas

briannosaurus:
Changed mongo mapreduce to pipe

Changelog: None

Forgot to add changelog, Ran test locally using go test in cmd folder. group returned the array backward from mapreduce but does not look like it is an issue

Signed-off-by: Brian Henshaw <brian@blkslabs.com>


alfrunes:
Change is no longer applicable with the new mongo-driver.
Please let us know if this issue is still present.

PR #285, opened 2019-06-07
cli: propagate-inventory

mchalski:
https://tracker.mender.io/browse/MEN-2479


cmd/commands.go
```diff
@@ -195,5 +204,140 @@ func decommissioningCleanupExecute(db *mongo.DataStoreMongo, dbName string) erro
...
 	return nil
 }
+
+func idDataToInventoryAttrs(id map[string]interface{}) ([]cinv.Attribute, error) {
+	var out []cinv.Attribute
+
+	for k, v := range id {
+		a := cinv.Attribute{
+			Name:  k,
+			Scope: "deviceauth",
```
kjaskiewiczz: why the scope is "deviceauth" instead of "identity"?

mchalski: oops, fixed

PR #286, opened 2019-06-20
Qa 75 production gitlab pipeline

mzedel:


.gitlab-ci.yml
```diff
@@ -0,0 +1,219 @@
...
+  - mongo
+
+stages:
+  - test_prep_build
+  - test_prep
+  - test
+  - build
+  - publish
+
+before_script:
```
lluiscampos: Same comment as in `mender-artifact`. It is used only once

.gitlab-ci.yml
```diff
@@ -0,0 +1,219 @@
...
+    # fail if any have been modified.
+    - if [ -n "$(go fmt)" ]; then echo 'Code is not formatted with "go fmt"'; false; fi
+    # Perform static code analysys
+    - go vet `go list ./... | grep -v vendor`
+    # Fail builds when the cyclomatic complexity reaches 15 or more
+    - gocyclo -over 15 `find . -iname '*.go' | grep -v 'vendor' | grep -v '_test.go'`
+    # Verify that the Swagger docs are valid
+    - swagger-cli validate docs/*.yml
+    # Verify that the Swagger docs follow the autodeployment requirements
+    - if test "$(ls -A docs)"; then python2.7 verify_docs.py `find docs -name "*.yml"`; fi
```
lluiscampos: Cleaner with something like: if test -d docs; then...

We can change to Python 3 here too after my  PR in autodocs

kacf:
I still don't get why code coverage isn't reported. Any clue, @mzedel?

mzedel:
unfortunately I have no idea... it should have been successfully reported here: https://travis-ci.org/mendersoftware/deviceauth/jobs/559491055#L3058 - I triggered yet another build, maybe it runs through this time...

kacf:
Indeed. Maybe it's getting confused with having both gitlab and travis available, or something. The results were in fact uploaded [here](https://codecov.io/github/mendersoftware/deviceauth/commit/d5914e620562be9b55e88646a6f6daf876fa255e), but never showed up here.

Let's just merge this. We are not breaking anything on the code level here, and we can focus on fixing codecov after the Gitlab transition is done.

PR #290, opened 2019-07-30
removed superfluous response type definition in swagger docs

mzedel:
Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

### DIFF
--- a/docs/management_api.yml
+++ b/docs/management_api.yml
@@ -472,10 +472,9 @@ definitions:
     example:
       application/json:
         identity_data:
-          application/json:
-            mac: "00:01:02:03:04:05"
-            sku: "My Device 1"
-            sn:  "SN1234567890"
+          mac: "00:01:02:03:04:05"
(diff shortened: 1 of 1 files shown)


mchalski:
@mzedel we tried that before and it failed spectacularly for the rendered docs (or something - something along the pipeline failed competely). did anything change along the way?

docs/management_api.yml
```diff
@@ -472,10 +472,9 @@ definitions:
     example:
       application/json:
```
mzedel: @mchalski that seems to be required

mzedel:
removing the outer `application/json` seems to be problematic, this one is basically an outlier in the responses

docs/management_api.yml
mchalski: oh, oh... I can see it now. approved - if you're sure it renders ok .

PR #292, opened 2019-08-05
mongodb: create index on auth sets collection

kjaskiewiczz:
New index containing sha256 of identity data and public key created.
Devauth App is looking for auth set existance querying by sha256 of
identity data and public key.


kjaskiewiczz:
it's about creating one index; I didn't provide the migration test by purpose

PR #294, opened 2019-09-05
Add support for default tenant token.

kacf:
It can be used to allow devices that don't have a tenant token to be
allowed into a specific tenant's list of devices. Enable it either
using the `DEVICEAUTH_DEFAULT_TENANT_TOKEN` environment variable, or
the `default_tenant_token` setting in `config.yaml`.

MEN-2705
MEN-2706

Changelog: Commit

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -58,7 +58,7 @@
 # Defaults to: none
 # Overwrite with environment variable: DEVICEAUTH_TENANTADM_ADDR
 
-# device_auth_orchestrator:  http://tenantadm
+# tenantadm_addr:  http://tenantadm
 
 # Private key path - used for JWT signing
 # Defaults to: /etc/deviceauth/rsa/private.pem
--- a/config.yaml
+++ b/config.yaml
@@ -60,6 +60,12 @@
 
 # tenantadm_addr:  http://tenantadm
 
+# Default tenant token, for devices that don't supply one (optional)
+# Defaults to: none
+# Overwrite with environment variable: DEVICEAUTH_DEFAULT_TENANT_TOKEN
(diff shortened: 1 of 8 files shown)


kacf:
My first stab at a backend feature.

config.yaml
```diff
@@ -58,7 +58,13 @@
...
 # Overwrite with environment variable: DEVICEAUTH_TENANTADM_ADDR
 
-# device_auth_orchestrator:  http://tenantadm
+# tenantadm_addr:  http://tenantadm
+
+# Default tenant token, for devices that don't supply one (optional)
+# Defaults to: none
+# Overwrite with environment variable: DEVICEAUTH_DEFAULT_TENANT_TOKEN
+
+# default_tenant_token:  abcdef
```
kjaskiewiczz: I'm not sure about this comment
I'm afraid it may suggest that the default_tenant_token may be anything, but the default_tenant_token needs to be a valid tenant token (valid jwt token with tenant claim)

kacf: Good point, I will replace it with:

```
# default_tenant_token:  <VALID_TENANT_TOKEN>
```

kacf:
Sorry, @kjaskiewiczz, there was yet another bugfix in https://github.com/mendersoftware/deviceauth/pull/294/commits/d8cc17ea9b6c09c03e5e916f2d398f1216654c77! Could you take a look? :-)

kjaskiewiczz:
looks good

kacf:
Thanks!

PR #296, opened 2019-09-12
Return device id to a POST /devauth/devices call

oleorhagen:
Changelog: Title

Signed-off-by: Ole Petter <ole.orhagen@northern.tech>

### DIFF
--- a/api/http/api_devauth.go
+++ b/api/http/api_devauth.go
@@ -211,6 +211,7 @@ func (d *DevAuthApiHandlers) PostDevicesV2Handler(w rest.ResponseWriter, r *rest
 	err = d.devAuth.PreauthorizeDevice(ctx, reqDbModel)
 	switch err {
 	case nil:
+		w.Header().Set("Location", "devices/"+reqDbModel.DeviceId)
 		w.WriteHeader(http.StatusCreated)
 	case devauth.ErrDeviceExists:
 		rest_utils.RestErrWithLog(w, r, l, err, http.StatusConflict)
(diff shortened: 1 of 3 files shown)


kjaskiewiczz:
it would be better to use standard "Location" header; value should refer to the resource created by the request (like here: https://docs.mender.io/2.0/apis/management-apis/user-administration-and-authentication#create-user)
see https://tools.ietf.org/html/rfc7231#section-7.1.2 for details

vendor/github.com/mendersoftware/go-lib-micro/testing/response.go
```diff
@@ -116,3 +116,25 @@ func (j *JSONResponse) CheckBody(t *testing.T, recorded *test.Recorded) {
 		assert.Empty(t, recorded.Recorder.Body.String())
 	}
 }
+
+type DevicePreauthReturnID struct {
```
kjaskiewiczz: you don't need custom checker (see comment above)
but even if you'll need you shouldn't modify packages from vendor directory directly
if you need to change go-lib-micro, you should open a PR to go-lib-micro repository, and when it will be merge you should open PR to this repo containing go-lib-micro update

oleorhagen: The reason the custom checker was added was because the ID generated is different from each time.

oleorhagen:
@kjaskiewiczz integration_master never finished. Or does it still not report?
The `test_full_integration/master` failed with: `https://github.com/mendersoftware/mender-artifact/pull/225`. Which I would label as spurious. Agree?

PR #297, opened 2019-09-24
devauth: fix context handling in ProvisionTenant

mchalski:
MigrateTenant always migrates the default db.

the main cause is - we're not setting up the context correctly with the
tenant's id.

actually, it works in part on the default db, in part on the tenant's db
(we use straight dbname in places, and context in others).

causes extremely difficult bugs when e.g. migrating from os to
enterprise (tenant's db needs migrating, so we try - but data
is taken from the default db, so it fails on dup keys).

changelog: none

Signed-off-by: Marcin Chalczynski <m.chalczynski@gmail.com>

### DIFF
--- a/devauth/devauth.go
+++ b/devauth/devauth.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


mchalski:
tests are hanging for some reason, but the green pipeline is already there: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/84105281

@kacf could you force merge?

PR #303, opened 2019-10-22
Update openssl lib version used in unit tests

lluiscampos:
It seems like a new package has been released upstream so the previous
one is not available anymore.

Changelog: None

Signed-off-by: Lluis Campos <lluis.campos@northern.tech>
(cherry picked from commit a92cf5ae2c81766a070c18481c8003fe2408dd3f)


lluiscampos:
@kacf FYI: I forgot to cherry-pick this into the release branch. It is not a blocker, but if for some other reason we need to do a `-build2` release candidate, please get this commit in.

lluiscampos:
Forgot about this one for a very long time. Merging now.

PR #307, opened 2020-01-10
Update golang version for unit tests to 1.11

lluiscampos:
Update golang version for unit tests to 1.11; same version as we use for the actual build.


lluiscampos:
Integration pipeline stopped.

PR #308, opened 2020-01-28
api/http: allow verification requests also on GET requests

mzedel:
Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>


mzedel:
Since the handler is exactly the same in both cases (GET/ POST), I didn't duplicate all the related test cases but only a single one to check if the GET access works...

mzedel:
one thing @mchalski: this might have to be included in the rate limiting efforts right?

PR #311, opened 2020-02-21
Fix: Not clearing unused JWT tokens from mongo in all cases

alfrunes:
Replaced the insert operation with an upsert. No more than a single device token should be active at the time.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,7 @@ variables:
   DOCKER_HOST: tcp://docker:2375/
 
 services:
-  - docker:dind
+  - docker:19.03-dind
   - mongo
 
 stages:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,7 @@ variables:
   DOCKER_HOST: tcp://docker:2375/
 
 services:
-  - docker:19.03-dind
+  - docker:19.03.5-dind
   - mongo
 
 stages:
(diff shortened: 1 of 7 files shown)


merlin-northern:
`TestStoreUpsertToken` and test licenses are failing.

alfrunes:
@merlin-northern the issues are solved now.

alfrunes:
> @alfrunes what will happen to existing tokens in the database?
> will they still lay around until we (manually) remove them?

Correct.
Perhaps it would be better to simply clear all existing tokens and replace the existing index with a TTL-index?

tranchitella:
@alfrunes what do you mean with TTL-index?

alfrunes:
@tranchitella
> TTL indexes are special single-field indexes that MongoDB can use to automatically remove documents from a collection after a certain amount of time or at a specific clock time.

https://docs.mongodb.com/v3.6/core/index-ttl/

tranchitella:
@alfrunes uh, I didn't know about TTL indexes, good spot!

No need to remove the other indexes, it seems. We would have to store the expiration date/time of the JWT token into a document property and index it with this `expireAfterSeconds` setting. Right? Sounds a very good approach to manage our JWT tokens.

merlin-northern:
@mirzak do you have any opinion on this one? this changes the behaviour that has been here for a long time, and means that we do not issue a new token for every auth set, but keep always one. I wonder if there is anything in the wild that relies on this?
I personally think it is a good change, but have some doubts about possible outcomes.

alfrunes:
Superseded by #318

PR #315, opened 2020-03-11
Feature/dependabot

mzedel:


oleorhagen:
The vendor folder has to be updated also:
```bash
go mod vendor
```

mzedel:
@oleorhagen wouldn't this switch to go modules entirely? if that's ok, I'll happily make the step - otherwise I'll leave that to another time...

oleorhagen:
> @oleorhagen wouldn't this switch to go modules entirely? if that's ok, I'll happily make the step - otherwise I'll leave that to another time...

It would. Fine with me. But I guess @tranchitella should have the final word here?

mzedel:
closing for #350

PR #316, opened 2020-03-24
[WiP] Migrate token collection model with TTL index to remove expired tokens

alfrunes:
This PR adreses the issue with cleaning up expired tokens using a TTL index. I've also enforced the `_id` unique index to be indexed by the authset id - thus no more than one active token per auth set. While I was at it I created a bson-compatible UUID type (see mendersoftware/go-lib-micro#55) which is more efficient to store and lookup [1]. The new token model preserves the old values, but ads the `IssuedAt` claim, which is required by the TTL index. The `Subject` claim still points to the device ID, but I also changed the datatype from string to binary UUID.

[1] https://docs.mongodb.com/manual/core/document/#the-id-field


alfrunes:
Superseded by #318

PR #318, opened 2020-04-01
TTL token migration

alfrunes:
The mongo datastore now store the token by payload/claims. The token id
`json:jti / bson:_id` now contains the authset id and the device id is 
contained in the jwt Subject (sub) claim. This implies that each device can
at most have one active token at the time.
Finally the token collection is indexed by a special TTL index on the ExpireAt
field of the token such that mongo can take care of removing expired tokens.


tranchitella:
Looks good! I've restarted the `test_full_integration/master` job as I believe it failed because of a glitch.

PR #323, opened 2020-04-10
JWT token verification: redis cache

tranchitella:


go.mod
```diff
@@ -0,0 +1,44 @@
+module github.com/mendersoftware/deviceauth
+
+go 1.13
+
+require (
```
alfrunes, line 5: Maybe we should add `go mod vendor` so that you don't need to pull all requirements on `go build`.

tranchitella: > Nice! Have you tried comparing the I/O throughput using Redis vs Mongo?
> This looks good to me, I only have a short comment regarding the `go mod` migration.

On my local machine, with a single device and docker-compose, using Redis as cache makes token verification 20% faster. The real advantage is with big databases (real world scenario), especially if you consider Redis will run in Kubernetes while mongodb is hosted. In those scenarios, I expect at least 50% faster response times and a huge difference in mongodb load (which is the biggest issue we have nowadays). It will also reduce data traffic (which we pay) between Kubernetes and Atlas.

tranchitella: Regarding the mod vendor, to be honest I've done it just because I failed to govendor the redis client library. I don't know what we'll decide about `go mod vendor`, we'll keep it consistent with other packages. :)

mchalski:
just for my information: what's the backstory on this work, any tasks or epics I can catch up on?

tranchitella:
@mchalski it's a PoC, no task/epic around, I would like to discuss it in one of the next back-end team meetings. Not urgent at all.

mchalski:
cool, just having redis is a good direction in general. I regret never been able to really go crazy with it, but the amount of usecases it can serve is so impressive. so +1.

I like to come back to this inspiring post http://oldblog.antirez.com/post/take-advantage-of-redis-adding-it-to-your-stack.html, and promise myself that one day I'll solve all of my problems along these lines ;)

anyway, I find redis to be great at giving you solutions to problems you didn't think you have, but you do, and now they're trivial to solve. I'm sure it will come in handy more and more.

alfrunes:
@tranchitella I just realized that this PR is not on top of `mendersoftware:ttl-token-migration` branch, so the tokens don't expire in the DB (just a heads up).

tranchitella:
@alfrunes thanks for pointing it out! When I said the token expires because of TTL, I meant expiration of the Redis key, which is independent from the TTL index in mongo:
https://github.com/mendersoftware/deviceauth/pull/323/commits/928b69685adaa16b42c56c612dc207cf1050eb93#diff-6cab6bba891bc37115b185fc2e565fbaR432

tranchitella:
Superseded by https://github.com/mendersoftware/deviceauth/pull/349, closing this PoC.

PR #324, opened 2020-04-14
MEN-3240 propagate device status to inventory: use the wf

merlin-northern:
Propagate device status to inventory
--

* **deviceauth** calls _update_device_status_ [[1](https://github.com/mendersoftware/workflows/pull/40)] workflow to propagate information on device status to inventory.
* **inventory** exposes internal API to set status for a list of devices [[2](https://github.com/mendersoftware/inventory-enterprise/pull/24)].
* _/filters/search_ can be use by UI to get list of devices as a single place of information.

related PRs:
 * [1] https://github.com/mendersoftware/workflows/pull/40
 * [2] https://github.com/mendersoftware/inventory/pull/162
 * [3] https://github.com/mendersoftware/deviceauth/pull/324

ChangeLog:none
Signed-off-by: Peter Grzybowski <peter@northern.tech>

### DIFF
--- a/api/http/api_devauth.go
+++ b/api/http/api_devauth.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Northern.tech AS
+// Copyright 2020 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
--- a/client/orchestrator/client_orchestrator.go
+++ b/client/orchestrator/client_orchestrator.go
@@ -32,6 +32,7 @@ const (
 	// orchestrator endpoint
 	DeviceDecommissioningOrchestratorUri = "/api/v1/workflow/decommission_device"
 	ProvisionDeviceOrchestratorUri       = "/api/v1/workflow/provision_device"
+	UpdateDeviceStatusOrchestratorUri    = "/api/v1/workflow/update_device_status"
 	// default request timeout, 10s?
 	defaultReqTimeout = time.Duration(10) * time.Second
 )
@@ -58,6 +59,19 @@ type ProvisionDeviceReq struct {
 	Device model.Device `json:"device"`
 }
 
+// UpdateDeviceStatusReq contains request data of request to start update
+// device status  workflow
+type UpdateDeviceStatusReq struct {
+	// Request ID
+	RequestId string `json:"request_id"`
+	// Device IDs
+	Ids string `json:"device_ids"`
+	// Tenant ID
+	TenantId string `json:"tenant_id"`
+	// new status
+	Status string `json:"device_status"`
+}
+
 // Config conveys client configuration
 type Config struct {
 	// Orchestrator host
@@ -70,6 +84,7 @@ type Config struct {
 type ClientRunner interface {
 	SubmitDeviceDecommisioningJob(ctx context.Context, req DecommissioningReq) error
 	SubmitProvisionDeviceJob(ctx context.Context, req ProvisionDeviceReq) error
+	SubmitUpdateDeviceStatusJob(ctx context.Context, req UpdateDeviceStatusReq) error
 }
 
 // Client is an opaque implementation of orchestrator client. Implements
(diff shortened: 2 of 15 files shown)


merlin-northern:
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/136596505

api/http/api_devauth.go
```diff
@@ -52,6 +52,8 @@ const (
 	v2uriToken               = "/api/management/v2/devauth/tokens/:id"
 	v2uriDevicesLimit        = "/api/management/v2/devauth/limits/:name"
 
+	v3uriDeviceStatus = "/api/management/v3/devauth/devices/:status"
```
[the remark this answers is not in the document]
merlin-northern: that is just an extra not-used-yet thing, I just added it since I believe this is the right way to proceed, and I was touching this anyway.

mchalski:
@merlin-northern I guess i need some fundamental clarifications:
- so the endpoint accepts a list of authsets - and it's public, how is this actually helping status propagation? my understanding was that on every status update devauth will post an update to inventory, why the public enpoint?
- related - I see you operate in bulk, on multiple IDs, can you clarify the scenario?

merlin-northern:
> @merlin-northern I guess i need some fundamental clarifications:
> 
> * so the endpoint accepts a list of authsets - and it's public, how is this actually helping status propagation? my understanding was that on every status update devauth will post an update to inventory, why the public enpoint?
> * related - I see you operate in bulk, on multiple IDs, can you clarify the scenario?

ok I can see that I have created a distraction and confusion with the extra (unused) endpoint I added. I have removed it, since it does not strictly belong to this PR.
all the update happens via workflow. I have also change the target branch to master.

client/orchestrator/client_orchestrator.go
```diff
@@ -58,6 +59,20 @@ type ProvisionDeviceReq struct {
 	Device model.Device `json:"device"`
 }
 
+// UpdateDeviceStatusReq contains request data of request to start update
+// device status  workflow
+type UpdateDeviceStatusReq struct {
+	// Request ID
+	RequestId string `json:"request_id"`
+	// Device IDs
+	//Ids []string `json:"device_ids"`
```
mchalski, line 68: which is it then? what's up with the list of strings?

client/orchestrator/client_orchestrator.go
```diff
@@ -58,6 +59,20 @@ type ProvisionDeviceReq struct {
...
+	Ids string `json:"device_ids"`
```
mchalski, line 69: it's either a single Id, or a list of Ids.

what's the backstory? in what scenario is it useful to update the status of multiple devices?

merlin-northern: the back story is: for now we only update a single device. but we have to change the UI to update multiple devices at once. so the workflow supports that, but you have to pass to the workflow a valid list in json, since there is no place to Marshal the arguments in the workflows, this has to be Marshalled [here](https://github.com/mendersoftware/deviceauth/pull/324/commits/d55f727cfd92f5ed7328cab221616abc57c9d03c#diff-bafc5a98b88ba2baa09868cc001b23e0R473)
So the workflow is as follows:
1. device changes state
2. deviceauth creates one element array with the device id, marshals it into json string
3. deviceauth calls update_device_status workflow which passes the string from 2. containing json array of ids as a workflow [parameter](https://github.com/mendersoftware/inventory/pull/162/commits/1c00753a07a513366699ed866638060a788267aa#diff-d38bb1bb13c34ab0a761f45bb70cd491R623) into POST call to internal inventory [endpoint](https://github.com/mendersoftware/inventory/pull/162/commits/1c00753a07a513366699ed866638060a788267aa#diff-d38bb1bb13c34ab0a761f45bb70cd491R623).
At one point the device auth will have to be called from UI with multiple devices to update (that is what the endpoint I had added here was supposed to be doing.)

client/orchestrator/client_orchestrator.go
merlin-northern: see below, I will remove the commented line.

PR #328, opened 2020-04-23
Remove DEVICEAUTH_MAX_DEVICES_LIMIT_DEFAULT configuration option

alfrunes:
The configuration option is considered redundant and is overridden in
enterprise environments.


tranchitella:
> it is still present on production:
> 
> ```shell
>  65         - name: DEVICEAUTH_MAX_DEVICES_LIMIT_DEFAULT
>  66           value: "500"
> ```
> 
> so I cant approve that until it is clear that we are removing it from saas

We target staging, and `saas:staging` doesn't have this setting anymore.

PR #331, opened 2020-05-04
Updated uuid package in go-lib-micro to generic objectIDs

alfrunes:
I recently discovered that at some point we were using bson-type ObjectIDs, which is not compatible with the uuid implementation from previous implementation, so I updated the uuid library to transparently cover objectIDs and arbitrary strings as well to ensure that we don't break old IDs with the migration (mendersoftware/go-lib-micro#60).


alfrunes:
Started an integration pipeline manually here https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/143717069

tranchitella:
@alfrunes why are you targeting the `ttl-token-migration` feature branch?

alfrunes:
> @alfrunes why are you targeting the `ttl-token-migration` feature branch?

@tranchitella this feature branch was never merged into master because of this issue with device id claim having different data types for previous version of the service. With this change I believed I've solved the issue, which is part of the reason I changed the package in `go-lib-micro`.
Besides, I don't think master is using the `go-lib-micro/uuid` package yet.

PR #335, opened 2020-05-06
MEN-3506 Porpagate status: migration 1.7.0 and set force migration ve…

merlin-northern:
…rsion

Migration is added:
 * 1.7.0 version bump
 * for automate upgrades

propagate-inventory-statuses command now supports force-set-migration
 * allows to set the migration version in the db

Misc:
 * store StoreMigrationVersion added

ChangeLog:none
Signed-off-by: Peter Grzybowski <peter@northern.tech>

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -22,6 +22,7 @@ import (
 	"github.com/mendersoftware/go-lib-micro/config"
 	"github.com/mendersoftware/go-lib-micro/identity"
 	"github.com/mendersoftware/go-lib-micro/log"
+	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
 	mstore "github.com/mendersoftware/go-lib-micro/store"
 	"github.com/pkg/errors"
 
@@ -223,7 +224,7 @@ func PropagateInventory(db store.DataStore, c cinv.Client, tenant string, dryrun
 	return nil
 }
 
-func PropagateStatusesInventory(db store.DataStore, c cinv.Client, tenant string, dryRun bool) error {
+func PropagateStatusesInventory(db store.DataStore, c cinv.Client, tenant string, migrationVersion string, dryRun bool) error {
 	l := log.NewEmpty()
 
 	dbs, err := selectDbs(db, tenant)
@@ -231,16 +232,18 @@ func PropagateStatusesInventory(db store.DataStore, c cinv.Client, tenant string
 		return errors.Wrap(err, "aborting")
 	}
 
+	var errReturned error
 	for _, d := range dbs {
-		err := tryPropagateStatusesInventoryForDb(db, c, d, dryRun)
+		err := tryPropagateStatusesInventoryForDb(db, c, d, migrationVersion, dryRun)
 		if err != nil {
+			errReturned = err
 			l.Errorf("giving up on DB %s due to fatal error: %s", d, err.Error())
 			continue
 		}
 	}
 
 	l.Info("all DBs processed, exiting.")
-	return nil
+	return errReturned
 }
 
 func selectDbs(db store.DataStore, tenant string) ([]string, error) {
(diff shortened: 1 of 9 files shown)


merlin-northern:
I am not very proud of StoreMigrationVersion but I had to add it for the force-set-migration sake.

cmd/commands.go
```diff
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
```
tranchitella, line 20: Can you please move this import lower, in the "non std-lib" import group?

cmd/commands.go
```diff
@@ -374,16 +377,32 @@ func tryPropagateStatusesInventoryForDb(db store.DataStore, c cinv.Client, dbnam
...
 		err = updateDevicesStatus(ctx, db, c, tenant, status, dryRun)
 		if err != nil {
 			l.Infof("Done with DB %s status=%s, but there were errors: %s.", dbname, status, err.Error())
+			errReturned = err
 		} else {
 			l.Infof("Done with DB %s status=%s", dbname, status)
 		}
 	}
+	if migrationVersion != "" && !dryRun {
+		if err != nil {
```
tranchitella, line 391: if the last updateDeviceStatus returns nil as error, you'll miss this if.
Are you sure here you want to check err instead of errReturned?

cmd/commands_test.go
```diff
@@ -16,6 +16,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
```
tranchitella, line 19: Same here, please move the import down.

store/datastore.go
```diff
@@ -17,6 +17,7 @@ package store
 import (
 	"context"
 	"errors"
+	"github.com/mendersoftware/go-lib-micro/mongo/migrate"
```
tranchitella, line 20: Same here.

store/mongo/migration_1_7_0.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+type migration_1_7_0 struct {
+	ms  *DataStoreMongo
+	ctx context.Context
+}
+
+const (
+	migrationContextTimeouts = 360
+	devicesBatchSize         = 512
```
tranchitella, line 41: I love these 2^ numbers :)

store/mongo/migration_1_7_0.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/mendersoftware/deviceauth/model"
+)
+
+type migration_1_7_0 struct {
+	ms  *DataStoreMongo
+	ctx context.Context
+}
+
+const (
+	migrationContextTimeouts = 360
```
tranchitella, line 40: 360 seconds to set up to 512 device statuses? Maybe we can lower it, I believe 60 seconds are already too much!

store/mongo/migration_1_7_0.go
```diff
@@ -0,0 +1,115 @@
...
+			Name:       &indexAuthSet_IdentityDataSha256_PubKey,
+			Unique:     &_true,
+		},
+	}
+	cAuthSets := m.ms.client.Database(ctxstore.DbFromContext(m.ctx, DbName)).Collection(DbAuthSetColl)
+	authSetIndexes := cAuthSets.Indexes()
+	_, err := authSetIndexes.CreateOne(m.ctx, authSetUniqueIndex)
+	if err != nil {
+		return errors.Wrap(err, "failed to create index containing IdDataSha256 and PubKey on auth sets")
+	}
```
tranchitella, line 108: I think I already asked you this question in a different PR: is this index related to this migration?

merlin-northern: so sorry about that! I copied the same piece of code, with apparently the same comment of yours. thanks!

store/mongo/migration_1_7_0.go
merlin-northern: sure.-> 32 the way you like it :)

store/mongo/migration_1_7_0.go
merlin-northern: same here! :)

store/datastore.go
merlin-northern: sure.thanks. corrected.

cmd/commands_test.go
merlin-northern: sure thanks corrected

cmd/commands.go
merlin-northern: good catch! thanks! fixed

cmd/commands.go
merlin-northern: sure thanks corrected

PR #337, opened 2020-05-10
Update migrate command to be consistent with server --automigrate

tranchitella:
The migrate command, when used without the `--tenant` option, migrates
the database of all the tenants. This behaviour is consistent with the
`--automigrate` option of the `server` command.

If the `--tenant` option is specified, the migration runs only for the
provided tenant ID.

Changelog: none

### DIFF
--- a/cmd/commands.go
+++ b/cmd/commands.go
@@ -63,16 +63,20 @@ func Migrate(c config.Reader, tenant string, listTenantsFlag bool) error {
 
 	db = db.WithAutomigrate().(*mongo.DataStoreMongo)
 
-	tenantCtx := identity.WithContext(context.Background(), &identity.Identity{
-		Tenant: tenant,
-	})
-
-	dbname := mstore.DbFromContext(tenantCtx, mongo.DbName)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/commands.go
```diff
@@ -63,16 +63,25 @@ func Migrate(c config.Reader, tenant string, listTenantsFlag bool) error {
...
+	ctx := context.Background()
+	if tenant == "" {
+		err = db.Migrate(ctx, mongo.DbVersion)
+	} else {
+		tenantCtx := identity.WithContext(ctx, &identity.Identity{
+			Tenant: tenant,
+		})
+
+		dbname := mstore.DbFromContext(tenantCtx, mongo.DbName)
+		if err != nil {
```
merlin-northern, line 79: where does `err` come from in here? `DbFromContext` returns string, we could check if it is `""` but looking at the code it could only happen if mongo.DbName is "".

tranchitella: uh, interesting, you are right... this comes from the code already there (I just made it conditional with an if). Fixed.

PR #342, opened 2020-05-15
Align ttl-token-migration branch with master

alfrunes:
Changed the TTL migration version for the token collection to resolve the version conflict with master.
I also merged this branch with current master.


tranchitella:
@alfrunes wouldn't it better to rebase on top of master? Or is it a nightmare?

alfrunes:
> @alfrunes wouldn't it better to rebase on top of master? Or is it a nightmare?

You're right, I'll give it a go.

alfrunes:
Opened a new PR doing a rebase instead #343

PR #347, opened 2020-05-26
ratelimit overrides for tenant-wide settings

mchalski:
https://tracker.mender.io/browse/MEN-3551

tenant-wide ratelimits can be overriden, per device.

overrides will be soon inspected/compared in token verification, i.e. around here: https://github.com/mendersoftware/deviceauth/blob/master/devauth/devauth.go#L996

I'm reusing the same generic structures as https://github.com/mendersoftware/tenantadm/pull/153 - I'll move them to go-lib-micro as a next step.

### DIFF
--- a/model/api_limits.go
+++ b/model/api_limits.go
@@ -1,0 +1,51 @@
+// Copyright 2020 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tranchitella:
@mchalski approved straight away, as the structure is the same you used in tenantadm.

PR #348, opened 2020-05-27
vendor: go-lib-micro update

mchalski:
use `go-lib-micro/ratelimits` common structs.


mchalski:
merging - full integration passed, but it's just the status widget that says 'pending'.

PR #349, opened 2020-06-01
ratelimits + token caching

mchalski:
https://tracker.mender.io/browse/MEN-3552

_This is a final version of a draft, updating the description_

To make a pretty long story short:
- implements token caching on verification for max performance
- implements throttling on token verification - based on tenant-wide api quotas and burst limits with device-level overrides 
- based on redis with a structured schema (keyed by tenant and device ids)
- if redis not configured - all of the above is disabled, degrades to ordinary implementation
- even if cache is configured - it's all best effort (don't fail whole requests on transient cache errors)

---
EDIT:
- updated with a few bugfixes after a round of manual testing - redis keying issues, timing issues (clock util for mocking), etc.
- added the nginx workaround discussed in the jira issue


mchalski:
@tranchitella thank you, and good points!

I'll set some timeouts on the client (`client.WithTimeout`) and make them configurable via our standard `config`. I missed it, and it should be just standard 'connection management hygiene'.

I see the point about real world connection concerns as well. it's true that in my mind, I was planning to enforce throttling without mercy ;) and exit on any errors that don't allow us to apply it effectively.

but sidestepping the cache in case of transient problems shouldn't hurt us, so I'll stick to your proposal. worst case we won't apply limits and fetch the token, but it will kick in at the very next opportunity, and the cache will fill up with the necessary stuff over the following requests.

alright, that helps a lot!

config.yaml
```diff
@@ -80,3 +80,40 @@
...
+# Defaults to: 0
+# Overwrite with environment variable: DEVICEAUTH_REDIS_DB
+
+# redis_db: "0"
+
+# Redis client timeout (sec)
+# Defaults to: 5
+# Overwrite with environment variable: DEVICEAUTH_REDIS_TIMEOUT_SEC
+
+# redis_timeout_sec: "5"
```
tranchitella, line 113: 5 seconds is a rather high value, and it is going to kill the backend if Redis becomes unresponsive.
What do you think about lowering it to 1 second?

mchalski: don't mind it at all, I'll set it to 1 sec.

it makes sense since we're best effort and want to get out quickly in case of problems.

PR #317, opened 2020-03-30
Migrate to go modules (PoC)

merlin-northern:
to migrate:
* initialize go mod:
  * `go mod init github.com/mendersoftware/deviceauth`
* update vendor:
  * `rm -Rf vendor`
  * `go mod vendor`
* add mendertesting to go.mod:
  * `github.com/mendersoftware/mendertesting v0.0.0-20180410095158-9e728b524c29`
  * `chmod 755 vendor/github.com/mendersoftware/mendertesting/check_license_go_code.sh`
* `go build` to check

PR #350, opened 2020-06-02
vendor: Migrate to `go mod` for tracking vendored dependencies

alfrunes:


mzedel:
shall I merge https://github.com/mendersoftware/deviceauth/pull/315 and then you merge this one? or would you just get the dependabot config in here? + there also is #317

alfrunes:
> One question: I think the gitlab file needs a bump in the go version as well (`tests:static` and `tests:unit` jobs).
> I was moving to 1.13, which is the default on gitlab so I just removed the `image` tags, but imo 1.14 needs to be explicitly stated.

Good point, I'll bump the golang version to 1.14.

alfrunes:
> shall I merge #315 and then you merge this one? or would you just get the dependabot config in here? + there also is #317

That's up to you, I suppose I could just cherry pick that config file here if you want?
I didn't notice #317, but if @merlin-northern approves maybe we could continue with the migration here?

merlin-northern:
> > shall I merge #315 and then you merge this one? or would you just get the dependabot config in here? + there also is #317
> 
> That's up to you, I suppose I could just cherry pick that config file here if you want?
> I didn't notice #317, but if @merlin-northern approves maybe we could continue with the migration here?

no problem -- what ever is more comfortable! it is fine by me whatever you decide.

mzedel:
> > shall I merge #315 and then you merge this one? or would you just get the dependabot config in here? + there also is #317
> 
> That's up to you, I suppose I could just cherry pick that config file here if you want?
> I didn't notice #317, but if @merlin-northern approves maybe we could continue with the migration here?

yes, please cherry-pick the config and merge it all in one swoop 🎉

PR #357, opened 2020-06-16
cache: clear device JWTs on device rejection scenarios

mchalski:
https://tracker.mender.io/browse/MEN-3631
https://tracker.mender.io/browse/MEN-3632


mchalski:
@tranchitella a followup would be good, I'll prepare it.

PR #381, opened 2020-08-06
index on id data sha256, not id data

mchalski:
prevents 'index key too large' errors.

https://tracker.mender.io/browse/MEN-3816

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	DbVersion     = "1.8.0"
+	DbVersion     = "1.9.0"
 	DbName        = "deviceauth"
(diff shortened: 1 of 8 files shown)


store/mongo/datastore_mongo.go
```diff
@@ -428,6 +429,10 @@ func (db *DataStoreMongo) MigrateTenant(ctx context.Context, database, version s
 			ds:  db,
 			ctx: ctx,
 		},
+		&migration_1_9_0{
```
alfrunes, line 432: Wouldn't it be reasonable to exclude `migration_1_1_0` since you're basically undoing it here?

mchalski: hmm we could in this case, but on principle I don't touch old migrations;
it's so easy to screw sth up for some particular user of a particular version...

PR #382, opened 2020-08-07
MEN-3813: device preauthorization: in case of conflict return conflicting device

kjaskiewiczz:
Issue: https://tracker.mender.io/browse/MEN-3813

When the user is trying to preauthorize the device that is already in the system, the service will return an HTTP 409 Conflict with the conflicting device in the body.


tranchitella (approved):

kjaskiewiczz:
@tranchitella should we include this fix in the 2.5 release?

tranchitella:
@kjaskiewiczz yes, I think it makes sense, do you agree?

kjaskiewiczz:
yes, I'll prepare a PR

PR #370, opened 2020-07-04
MEN-3763 GetDevicesById

merlin-northern:
ChangeLog:none
Signed-off-by: Peter Grzybowski <peter@northern.tech>

PR #398, opened 2020-09-11
Extend GET /devices with 'id' query parameter to filter on specific devices

alfrunes:
Needed by MEN-3749 and MEN-3750 for the internal endpoint. (Alternative version to #370 for the purpose of this task)

NOTE: For the management endpoint there is a default header restriction of 4KiB in the gateway which limits the number of ids you can practically filter on, but should suffice for up to ~60-70 ids. As for the internal endpoint, Go http is a little more lax with 1MiB default max header size.


alfrunes:
@mzedel would this solve your [issue](https://github.com/mendersoftware/deviceauth/pull/370#issuecomment-684640324) as well (given the restriction mentioned above)?
If so I think we can close #370 in favor of this one (I'm not a huge fan of using POST for query requests).

mzedel:
Thanks @alfrunes, that should solve it as well... and the gateway limitation should be alright, as the maximum page length is 50 at the moment... we could probably get around it with https://github.com/mendersoftware/integration/pull/797 🤷‍♂️

lluiscampos:
Ups. I messed up the ongoing builds for this PR... Sorry. Retying :)

model/device.go
```diff
@@ -71,3 +83,35 @@ func NewDevice(id, id_data, pubkey string) *Device {
...
+			`parameter status must be one of: `+
+				`%s, %s, %s or %s`,
+			DevStatusPending, DevStatusRejected,
+			DevStatusAccepted, DevStatusPreauth,
+		)
+	}
+	return nil
+}
+
+func (fltr *DeviceFilter) ParseForm(form url.Values) error {
```
tranchitella, line 109: Adding a unit test here is easy and it increases the code test coverage

tranchitella (approved):

tranchitella (approved):

PR #406, opened 2020-09-30
client: Reuse tcp connections by not destroying http.Client objects

alfrunes:
Orchestrator and tenant clients are initializing a new http.Client on
every request, costing an additional tcp handshake for trivial requests.


tranchitella (approved):
The PR is fine, but can you please elaborate a bit the additional tcp handshake you described in the description? Does it mean the HTTP connection will be kept open between the different `Do` calls, if you reuse the same http Client? Is this something you tested?

I'm not a go expert, but my (probably wrong, at this point) impression was that each `httpClient.Do` call would start a new HTTP call, which brings into the equation the initial TCP handshake.

mchalski:
surely +1 for reusing, there's no need to recreate these clients constantly; I can also see some added timeouts here and there, +1 as well.

to be perfectly honest, I wasn't aware of the behavior you describe though. I would second Fabio's assumption immediately...

I brushed up a bit and assume now you're referring to this bit in the docs:
```
The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines.
```
and
```
DefaultTransport is the default implementation of Transport and is used by DefaultClient. It establishes network connections as needed and caches them for reuse by subsequent calls. 
```
indeed transports cache connections - even as far as per address etc. in `idleConn     map[connectMethodKey][]*persistConn // most recently used at end`.

it's a tentative approve from me then.

mchalski (approved):

alfrunes:
Sorry about the brief description.
I was just a little annoyed by the fact that we keep re-initializing a new client for each request. The http.Client.Transport (as @mchalski points out) struct contains a map of idle connections which it by default retains for 90 seconds after the connection becomes idle. By re-initializing the `http.Client`, I believe this map is created and destroyed as the client goes out of scope.

tranchitella:
@alfrunes are you following up on all the other services/repositories? If not, let's create a task for it, it seems a good optimization.

alfrunes:
@tranchitella Fortunately many of the clients are already following this pattern, but I'll create a ticket to be sure we'll make all client implementations consistent.

PR #407, opened 2020-10-05
MEN-3862: New command 'check-device-limits' for warning tenants approaching their device limits

alfrunes:
This PR adds a command that iterates all tenant DBs and checks if the number of devices accepted are above a threshold of the total device limit for the given tenant and fires a `device_limit_email` workflow if the assertion is true.


store/mongo/datastore_mongo.go
```diff
@@ -700,39 +700,21 @@ func (db *DataStoreMongo) GetLimit(ctx context.Context, name string) (*model.Lim
 }
 
 func (db *DataStoreMongo) GetDevCountByStatus(ctx context.Context, status string) (int, error) {
-	c := db.client.Database(ctxstore.DbFromContext(ctx, DbName)).Collection(DbDevicesColl)
+	var (
+		fltr     = bson.D{}
+		devsColl = db.client.
+				Database(ctxstore.DbFromContext(ctx, DbName)).
+				Collection(DbDevicesColl)
+	)
```
tranchitella, line 708: Well done!

tranchitella (approved):

PR #408, opened 2020-10-05
tests: Update docker image and python packages for acceptance tests

alfrunes:
Python 3.5 has already reached EOL so bumped the image to python 3.8.
Moreover, pycrypto seems to no longer be maintained and makes use of
deprecated function time.clock breaking compatibility with python 3.8
leading to the replacement with `cryptography` package. Finally, the
pymongo version was bumped to the latest stable as well to suppress some
deprecation warnings.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,6 +1,6 @@
-FROM python:3.5
+FROM python:3.8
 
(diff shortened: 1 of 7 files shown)


mzedel:
🎉 came across this in tenantadm today and made a mental note to finally fix this across the repos - so thank you for tackling it here!

tranchitella (approved):
Thanks @alfrunes, I really appreciate your attention to details.

PR #414, opened 2020-10-26
propagate device identity using workflows

kjaskiewiczz:
Issue: https://tracker.mender.io/browse/MEN-3979
When propagating device identity attributes replace point-to-point communication between deviceauth and inventory with workflows.
The "status" attribute name is reserved. If there will be an attribute called "status" in the identity attributes it will be ignored by the backend.


alfrunes (commented):
The PR seems fine, but all the deployment integration tests are suddenly failing with off-by-one errors for some reason...

kjaskiewiczz:
@alfrunes exactly! I've asked for the review to check if someone else can find some obvious issue with my PR, because I can't see one, and debugging the tests is taking ages

alfrunes:
@kjaskiewiczz I've done some manual testing and everything seems fine. The only thing I can think of is if there is a race condition between the update inventory workflow and starting a dynamic deployment in the tests.

kjaskiewiczz:
@alfrunes thanks for testing this; I was also testing this change manually and everything looks fine;
inventory (identity) update workflow shouldn't affect tests IMO, since the tests are using inventory parameters (with inventory scope), not identity
anyway, thanks again for the help and I'm going back to my investigation

devauth/devauth.go
```diff
@@ -221,7 +257,9 @@ func (d *DevAuth) getDeviceFromAuthRequest(ctx context.Context, r *model.AuthReq
 		tenantId = idData.Tenant
 	}
 	if addDeviceErr != store.ErrObjectExists {
-		d.invClient.SetDeviceIdentity(ctx, tenantId, dev.Id, dev.IdDataStruct)
+		if err := d.setDeviceIdentity(ctx, dev, tenantId); err != nil {
+			return nil, err
```
mchalski, line 261: err != nil ... return nil?

tranchitella:
Closing this old PR, as we addressed the issue in https://github.com/mendersoftware/deviceauth/pull/417

PR #417, opened 2020-11-08
propagate identity data using workflows + new device status synchronization flow

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-3979 https://tracker.mender.io/browse/MEN-4112


devauth/devauth.go
```diff
@@ -519,22 +557,27 @@ func (d *DevAuth) updateDeviceStatus(ctx context.Context, devId, status string,
 		return errors.Wrap(err, "failed to update device status")
 	}
 
-	b, err := json.Marshal([]string{devId})
-	if err != nil {
-		return errors.New("internal error: cannot marshal array into json")
-	}
 	// submit device status change job
 	if statusChanged {
+		dev, err := d.db.GetDeviceById(ctx, devId)
```
tranchitella, line 562: This extra GetDeviceById can probably be avoided using FindAndModify in SetDeviceIdentity.
It doesn't matter, as we do not expect to change statuses so often, let's keep it as it is.

tranchitella (approved):

devauth/devauth.go
kjaskiewiczz: you mean in the UpdateDevice, right?
I'll follow up on this, we also call GetDeviceStatus which is not always needed

kjaskiewiczz:
green pipeline - https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/216725608

PR #422, opened 2020-11-18
Merge master into feature-deviceconnect

tranchitella:


tranchitella:
Manually launched pipeline:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/217872219

alfrunes (approved):

tests/tests/orchestrator.py
```diff
@@ -25,11 +25,12 @@ def provision_device_handler(device_id=None, status=200):
     log = logging.getLogger("orchestartor.provision_device")
 
     def _provision_device(request):
```
alfrunes, line 27: I believe this is wrong, I updated the `provision_device` workflow parameter to contain a "device_id" property, not the entire Device

alfrunes: IOW; the diff should be the other way around.

tranchitella: RIght, fixed here: https://github.com/mendersoftware/deviceauth/pull/422/commits/a25729b4a119c48a79f583c9b3cb95607286dfdc

PR #423, opened 2020-11-20
Fix device status propagation

kjaskiewiczz:


tranchitella (approved):
Thanks a lot, it looks great!

tranchitella (changes requested):
Sorry, I just noticed that noauth status is missing here:
https://github.com/mendersoftware/deviceauth/blob/master/cmd/commands.go#L424

tranchitella (approved):
Thanks!

PR #424, opened 2020-11-21
Fix: return 401 unauthorized on ErrAuthSetNotFound

tranchitella:
In case an authset is not found when verifying a JWT token, we returned
500 internal server error. We change the behaviour to return 401
Unauthorized, letting the device to ask for a new JWT token.

Changelog: none


tranchitella:
Only `tests/test_mtls.py::TestClientMTLSEnterprise::test_mtls_enterprise_hsm[rsa]` failed in integration test.

alfrunes (approved):

PR #426, opened 2020-12-02
MEN-3763: New endpoint for querying devices with auth sets

alfrunes:
New endpoint `POST /api/management/v2/devauth/devices/search`

>    The endpoint acts as a POST form version of
>    GET /api/management/v2/devauth/devices
>    with the exact same parameters. This commit also allows querying over
>    both multiple and singular device statuses and ids to both endpoints.
>    The new POST endpoint accepts both 'application/json' and
>    'application/x-www-form-urlencoded' (the latter will not be enabled by
>    default middleware, however).

### DIFF
--- a/api/http/api_devauth.go
+++ b/api/http/api_devauth.go
@@ -53,6 +53,7 @@ const (
 	// management API v2
 	v2uriDevices             = "/api/management/v2/devauth/devices"
 	v2uriDevicesCount        = "/api/management/v2/devauth/devices/count"
+	v2uriDevicesSearch       = "/api/management/v2/devauth/devices/search"
 	v2uriDevice              = "/api/management/v2/devauth/devices/:id"
 	v2uriDeviceAuthSet       = "/api/management/v2/devauth/devices/:id/auth/:aid"
 	v2uriDeviceAuthSetStatus = "/api/management/v2/devauth/devices/:id/auth/:aid/status"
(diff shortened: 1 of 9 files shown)


api/http/api_devauth.go
```diff
@@ -254,27 +256,42 @@ func (d *DevAuthApiHandlers) PostDevicesV2Handler(w rest.ResponseWriter, r *rest
...
+		if err = fltr.ParseForm(r.Form); err != nil {
+			rest_utils.RestErrWithLog(w, r, l, err, http.StatusBadRequest)
+			return
+		}
+
+	default:
+		rest_utils.RestErrWithLog(w, r, l, errors.Errorf(
+			"Content-Type '%s' not supported",
+			r.Header.Get("Content-Type"),
+		), http.StatusBadRequest)
```
kjaskiewiczz, line 294: what about responding with 415 Unsupported Media Type

api/http/api_devauth.go
```diff
@@ -305,7 +322,12 @@ func (d *DevAuthApiHandlers) GetDevicesV2Handler(w rest.ResponseWriter, r *rest.
 		return
 	}
 
-	w.WriteJson(outDevs)
+	_ = w.WriteJson(outDevs)
```
tranchitella, line 325: How of curiosity, why do you need the `_ =`?

tranchitella (approved):

api/http/api_devauth.go
alfrunes: I have `golangci-lint` integrated in my editor, and the `errcheck` linter complains that the returned error is not handled.
This syntax indicates that I purposely ignore the error (there's nothing else we really can do in these situations).

tranchitella: Thanks for sharing, I'll install the golancgi-lint, too!

api/http/api_devauth.go
alfrunes: Actually, this code is unreachable because of the `rest.ContentTypeCheckerMiddleware` but fixed it regardless.

PR #427, opened 2020-12-02
MEN-3972: ignore tenant claim in single tenant setup

kjaskiewiczz:


devauth/devauth.go
```diff
@@ -381,6 +381,12 @@ func (d *DevAuth) SubmitAuthRequest(ctx context.Context, r *model.AuthReq) (stri
 		if err != nil {
 			return "", err
 		}
+	} else {
+		// ignore identity data when tenant verification is off
+		// it's possible that the device will provide old auth token or old tenant token
+		// in the authorization header;
+		// in that case we need to wipe identity data from the context
+		ctx = identity.WithContext(ctx, &identity.Identity{})
```
alfrunes, line 389:
```suggestion
		ctx = identity.WithContext(ctx, nil)
```
I think an empty `Identity` struct will cause the resolve Database name to become `"deviceauth-"`.

kjaskiewiczz: it's enough that the tenant id is empty - https://github.com/mendersoftware/go-lib-micro/blob/4c9081b6f34ee1c01b6ee394bcf12b7ec00a24e3/store/utils.go#L57 but still, nil looks probably better

alfrunes (approved):

PR #431, opened 2020-12-04
api: Add option to filter devices by id for GET /devices

tranchitella:
The aggregation pipeline was replaced by a Find operation on the storage
layer and the DeviceFilter was moved to the 'model' package.

changelog: New query parameter id for GET /api/management/v1/devices

Signed-off-by: Alf-Rune Siqveland <alf.rune@northern.tech>


alfrunes (approved):

tranchitella:
The only failed test in integration is unrelated to these changes. I'm merging.

PR #475, opened 2021-03-17
MEN-4512: Addon restrictions for devices

alfrunes:
Since there are no plan restrictions for devices (yet) I didn't implement that part, but I tried to keep the `access` package flexible with this in mind.

### DIFF
--- a/access/addons.go
+++ b/access/addons.go
@@ -1,3 +1,17 @@
+// Copyright 2021 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


alfrunes:

mchalski (approved):
looks good.

FTR I had some doubts about maybe also introducing `TestToken` open source - but it doesn't make sense. `deviceconfig` will not even be there in the open source setup...

(it did make sense to test both setups in useradm-enterprise - but only because they effectively translate to 'enterprise hosted' and 'enterprise onprem', so it was a good opportunity to check that addons are bypassed if !HAVE_ADDONS, which will be 99% of cases for onprem users, they're not interested in this at all. this is not the case here, but anywa, I'm rambling, approved :) )

alfrunes:
The integration test errors appears to be spurious. I tested the failing tests locally and obtained the following results:
 * `test_bogus_protomsg`: passed
 * `test_filetransfer`: passed
 * `test_configuration`: passed for open source test and failed for enterprise (the opposite of what happened in the pipeline).

alfrunes:

PR #485, opened 2021-04-06
MEN-4604: make sure the status is propagated to inventory for pending…

tranchitella:
… devices

If the backend is under load (many new devices authorizing), it can
happen the status attribute is not propagated to the inventory service.
This leads to a device not visible in the UI, but correctly provisioned
as pending in deviceauth.

To avoid it we update the flow of the auth request:

1. Default status for devices is noauth
2. Auth request processing happens as usual
3. The workflow execution to update the status in inventory is started
4. If no error happened in 1-3, we updated the status in db to pending

In case of failure, the status in the db will be noauth, which will
trigger a new sync to inventory because noauth != pending.

Changelog: none

### DIFF
--- a/devauth/devauth.go
+++ b/devauth/devauth.go
@@ -233,7 +233,7 @@ func (d *DevAuth) setDeviceIdentity(ctx context.Context, dev *model.Device, tena
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


tranchitella:

tranchitella:

tranchitella:

alfrunes (approved):

lluiscampos (approved):
Far from an expert in this area, but the fix looks good to me. Thanks for pinging me here :+1:

PR #500, opened 2021-06-05
Improve the invalidation of the JWT token cache to avoid race conditions

tranchitella:


tranchitella:

kjaskiewiczz (approved):
looks good

tranchitella:
Manual pipeline is fully green here:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/315719988

devauth/devauth.go
```diff
@@ -1076,23 +1070,22 @@ func (d *DevAuth) PreauthorizeDevice(ctx context.Context, req *model.PreAuthReq)
...
 	tokenOID := oid.FromString(tokenID)
 
-	if d.cache != nil {
-		token, err := d.db.GetToken(ctx, tokenOID)
+	l.Warnf("Revoke token with jti: %s", tokenID)
+	err := d.db.DeleteToken(ctx, tokenOID)
+
+	if err == nil && d.cache != nil {
+		var token *jwt.Token
+		token, err = d.db.GetToken(ctx, tokenOID)
```
mchalski, line 1081: @tranchitella I'm missing something here:

aren't we guaranteed to not get a token here (right after `err := d.db.DeleteToken(ctx, tokenOID)`, and never proceed to actually clearing the cache?

could you expand a bit on where the race condition occurs?

kjaskiewiczz: right; looks like we should get the token from db first, remove it from db and clear the cache

tranchitella: You are right, sorry, dumb mistake.

tranchitella:

mchalski (approved):

PR #535, opened 2021-09-20
[MEN-5077] Restrict access to monitor APIs to devices with monitor add-on

alfrunes:


tranchitella (approved):

tranchitella:
@alfrunes changelog is missing from the first commit, otherwise looks okay!

PR #536, opened 2021-09-22
[MEN-5077] Restrict access to monitor APIs to devices with monitor add-on

alfrunes:
Cherry pick #535


lluiscampos:
I already merged this manually - when doing minor release I sometimes fast-forward the release branches if there are no other changes in the given repository (as it was the case for `deviceauth`

PR #538, opened 2021-09-29
MEN-5124: do not hard-code the email sender, rely on workflows' default

tranchitella:
Changelog: none

### DIFF
--- a/client/orchestrator/client_orchestrator_test.go
+++ b/client/orchestrator/client_orchestrator_test.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tranchitella:
Combined QA pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/379260004

client/orchestrator/client_orchestrator_test.go
```diff
@@ -259,12 +258,11 @@ func TestSubmitDeviceLimitWarning(t *testing.T) {
 		Error: errors.New(
 			`workflows: \[internal\] invalid request argument: ` +
 				`invalid device limit request: ` +
-				`missing parameter "from"`),
+				`missing parameter "to"`),
```
merlin-northern, line 261: wait I do not get it, now "to" is missing?

merlin-northern (approved):
lgtm with some nitpick in tests

client/orchestrator/client_orchestrator_test.go
tranchitella: from is not mandatory anymore (that's the change)

PR #548, opened 2021-10-29
MEN-4843: add (optional) support to trigger the reporting service rei…

tranchitella:
…ndex

Chagelog: title


kjaskiewiczz (approved):
looks good

PR #551, opened 2021-11-14
QA-235: pipeline: Switch to static checks by golangci-lint

lluiscampos:
And fix the code accordingly.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -19,7 +19,8 @@
 //
 // 1. Quota enforcement
 //
-// Based on https://redislabs.com/redis-best-practices/basic-rate-limiting/, but with a flexible interval (ratelimits.ApiQuota.IntervalSec).
+// Based on https://redislabs.com/redis-best-practices/basic-rate-limiting/, but with a flexible
+// interval (ratelimits.ApiQuota.IntervalSec).
 // Current usage for a device lives under key:
 //
 // `tenant:<tid>:device:<did>:quota:<interval_num>: <num_reqs>`
@@ -48,9 +49,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/pkg/errors"
 	"strconv"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/go-redis/redis/v8"
 	"github.com/mendersoftware/go-lib-micro/ratelimits"
@@ -70,11 +72,21 @@ var (
 //go:generate ../utils/mockgen.sh
 type Cache interface {
 	// Throttle applies desired api limits and retrieves a cached token.
-	// These ops are bundled because the implementation will pipeline them for a single network roundtrip for max performance.
+	// These ops are bundled because the implementation will pipeline them for a single network
+	// roundtrip for max performance.
 	// Returns:
 	// - the token (if any)
 	// - potentially ErrTooManyRequests (other errors: internal)
-	Throttle(ctx context.Context, rawToken string, l ratelimits.ApiLimits, tid, id, idtype, url, action string) (string, error)
+	Throttle(
+		ctx context.Context,
+		rawToken string,
+		l ratelimits.ApiLimits,
+		tid,
+		id,
+		idtype,
+		url,
+		action string,
+	) (string, error)
 
 	// CacheToken caches the token under designated key, with expiration
 	CacheToken(ctx context.Context, tid, id, idtype, token string, expireSec time.Duration) error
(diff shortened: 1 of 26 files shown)


cache/cache.go
```diff
@@ -151,7 +179,7 @@ func (rl *RedisCache) Throttle(ctx context.Context, rawToken string, l ratelimit
 	}
 
 	// collect quota/burst control and token fetch results
-	tok, err := rl.checkToken(tokenGet, rawToken)
+	tok, _ := rl.checkToken(tokenGet, rawToken)
```
tranchitella, line 182: We should return the error not swallowing it:
```suggestion
	tok, err := rl.checkToken(tokenGet, rawToken)
	if err != nil {
		return "", err
	}
```

client/inventory/client.go
```diff
@@ -84,6 +94,7 @@ func (c *client) CheckHealth(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	defer rsp.Body.Close()
```
tranchitella, line 97: I've seen these in the other PRs: the missing differed call to Close is indeed a bug, well done.

main.go
```diff
@@ -237,20 +241,24 @@ func cmdMigrate(args *cli.Context) error {
...
+	)
 	if err != nil {
 		return cli.NewExitError(err, 6)
 	}
 	return nil
 }
 
 func cmdPropagateStatusesInventory(args *cli.Context) error {
-	db, err := mongo.NewDataStoreMongo(makeDataStoreConfig())
+	db, _ := mongo.NewDataStoreMongo(makeDataStoreConfig())
```
tranchitella, line 256:
```suggestion
	db, err := mongo.NewDataStoreMongo(makeDataStoreConfig())
	if err != nil {
		return err
	}
```

main.go
```diff
@@ -262,12 +270,12 @@ func cmdPropagateStatusesInventory(args *cli.Context) error {
 }
 
 func cmdPropagateIdDataInventory(args *cli.Context) error {
-	db, err := mongo.NewDataStoreMongo(makeDataStoreConfig())
+	db, _ := mongo.NewDataStoreMongo(makeDataStoreConfig())
```
tranchitella, line 273:
```suggestion
	db, err := mongo.NewDataStoreMongo(makeDataStoreConfig())
	if err != nil {
		return err
	}
```

store/mongo/datastore_mongo_cleanup.go
```diff
@@ -138,22 +137,25 @@ func (db *DataStoreMongo) DeleteBrokenAuthSets(dbName string) error {
...
+	// delete authsets for non-existent devices
 	for _, as := range authSets {
-		_, err := collAuthSets.DeleteOne(nil, model.AuthSet{Id: as})
+		_, err := collAuthSets.DeleteOne(context.TODO(), model.AuthSet{Id: as})
 		if err != nil {
 			return errors.Wrapf(err, "database %s, failed to delete authentication sets", dbName)
 		}
 		// Attempt to delete token (may have already expired).
-		collTokens.DeleteOne(nil, bson.M{"_id": as})
+		_, _ = collTokens.DeleteOne(context.TODO(), bson.M{"_id": as})
```
tranchitella, line 147: Define the context outside of the for loop and reuse it.
We usually use context.Background in this case, instead of context.TODO.

store/mongo/migration_1_7_0.go
```diff
@@ -56,7 +58,7 @@ func (m *migration_1_7_0) updateDevicesStatus(ctx context.Context, status string
 	i = 0
 	for cur.Next(ctx) {
 		var d model.Device
-		err = cur.Decode(&d)
+		_ = cur.Decode(&d)
```
tranchitella, line 61:
```suggestion
		err = cur.Decode(&d)
		if err != nil {
			return err
		}
```

store/mongo/migration_1_7_0.go
```diff
@@ -81,10 +83,10 @@ func (m *migration_1_7_0) updateDevicesStatus(ctx context.Context, status string
...
-	m.updateDevicesStatus(ctx, "accepted")
-	m.updateDevicesStatus(ctx, "pending")
-	m.updateDevicesStatus(ctx, "rejected")
-	m.updateDevicesStatus(ctx, "preauthorized")
+	_ = m.updateDevicesStatus(ctx, "accepted")
+	_ = m.updateDevicesStatus(ctx, "pending")
+	_ = m.updateDevicesStatus(ctx, "rejected")
+	_ = m.updateDevicesStatus(ctx, "preauthorized")
 
 	return nil
```
tranchitella, line 91:
```suggestion
	err := m.updateDevicesStatus(ctx, "accepted")
	if err == nil {
		err = m.updateDevicesStatus(ctx, "pending")
	}
	if err == nil {
		err = m.updateDevicesStatus(ctx, "rejected")
	}
	if err == nil {
		err = m.updateDevicesStatus(ctx, "preauthorized")
	}
	return err
```

store/mongo/migration_1_8_0.go
```diff
@@ -39,6 +39,7 @@ func (m *migration_1_8_0) Up(from migrate.Version) error {
 	collTokens = database.Collection(DbTokensColl)
 	idxView := collTokens.Indexes()
 	idxOpts := mopts.Index()
+	//nolint:staticcheck
 	idxOpts.SetBackground(false)
```
tranchitella, line 43: You can remove `SetBackground(false)`

tranchitella (approved):
A few comments and suggestions

store/mongo/datastore_mongo_cleanup.go
lluiscampos: Noted. I used `TODO` because is what the docs suggest to use instead of `nil`. I'll rework.

store/mongo/migration_1_7_0.go
lluiscampos: Nice, thanks.

lluiscampos:
@mender-test-bot start pipeline :four:

PR #574, opened 2022-02-18
refactor: do not store the public key in the device document

tranchitella:
Changelog: title

### DIFF
--- a/api/http/api_devauth_test.go
+++ b/api/http/api_devauth_test.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


tranchitella:

tranchitella:
@mender-test-bot start pipeline --pr deviceauth-enterprise/48

mzedel (changes requested):
I'm sorry to bring bad news, but the [gui code](https://github.com/mendersoftware/gui/blob/master/src/js/components/devices/device-details/authsets/authsetlistitem.js#L90-L105) uses the key... if you can provide a separate endpoint (+ maybe bump the api version of this one) we could likely still get the optimization benefits

MaciejTe (approved):
It looks good from my perspective. Pre approving (please make sure these changes do not collide with GUI, as Manuel mentioned). Thanks!

tranchitella:
> I'm sorry to bring bad news, but the [gui code](https://github.com/mendersoftware/gui/blob/master/src/js/components/devices/device-details/authsets/authsetlistitem.js#L90-L105) uses the key... if you can provide a separate endpoint (+ maybe bump the api version of this one) we could likely still get the optimization benefits

I think you are wrong: you are processing the pubkey field of the authset. Here I'm talking about the pubkey field of the device, which is not even returned from the API.

image

mzedel (approved):
Excellent, thanks for the clarification and sorry for not checking this in depth!

alfrunes (approved):

PR #575, opened 2022-02-23
MEN-4745 move the service to one db for all tenants

MaciejTe:
refactor: move the service to the one database

Changelog: title

Signed-off-by: Maciej Tomczuk <maciej.tomczuk@northern.tech>


store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,185 @@
...
+			Indexes: []mongo.IndexModel{
+				{
+					Keys: bson.D{{Key: "exp.time", Value: 1}},
+					Options: &mopts.IndexOptions{
+						ExpireAfterSeconds: &expireAfterSeconds,
+						Name:               &tokenExpirationIndexName,
+					},
+				},
+			},
+		},
```
alfrunes, line 78: I think we can leave the tokens collection as is. Although this will lead to all devices to become unauthorized on the first request after the migration. Otherwise, I think this collection needs to be treated separately.

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,185 @@
...
+	collectionsIndexes := map[string]struct {
+		Indexes []mongo.IndexModel
+	}{
+		DbDevicesColl: {
+			Indexes: []mongo.IndexModel{
+				// create device index on status and id
+				{
+					Keys: bson.D{
+						{Key: model.DevKeyId, Value: 1},
+						{Key: model.TenantIDField, Value: 1},
```
alfrunes, line 49: I think the TenantIDField should be the top-level key.

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,185 @@
...
+						{Key: model.AuthSetKeyDeviceId, Value: 1},
+						{Key: model.AuthSetKeyIdDataSha256, Value: 1},
+						{Key: model.AuthSetKeyPubKey, Value: 1},
+					},
+					Options: &options.IndexOptions{
+						Background: &_false,
+						Name:       &indexAuthSet_DeviceId_IdentityDataSha256_PubKey,
+						Unique:     &_true,
+					},
+				},
```
alfrunes, line 103: Uhh... Isn't this index completely redundant? The above index is unique and contains a subset of the keys...

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,185 @@
...
+				{
+					Keys: bson.D{
+						{Key: model.DevKeyId, Value: 1},
+						{Key: model.TenantIDField, Value: 1},
+					},
+					Options: &mopts.IndexOptions{
+						Background: &_false,
+						Name:       &indexDevices_Status,
+						Unique:     &_false,
+					},
```
alfrunes, line 55: The options have a cleaner functional interface :)
Also, the `Background` option is deprecated since mongo 4.2.
```suggestion
					Options: mopts.Index().
						SetName(indexDevices_Status).
						SetUnique(false),
```

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,185 @@
...
+						Name:               &tokenExpirationIndexName,
+					},
+				},
+			},
+		},
+		DbAuthSetColl: {
+			Indexes: []mongo.IndexModel{
+				{
+					Keys: bson.D{
+						{Key: model.AuthSetKeyIdDataSha256, Value: 1},
```
alfrunes, line 83: I think we should add the `tenant_id` as the top level key in case of collisions.

cmd/commands.go
```diff
@@ -463,95 +461,3 @@ func tryPropagateIdDataInventoryForDb(
...
-			}
-		}
-		return nil
-	}
-	// Start looping through the databases.
-	return ds.ForEachDatabase(
-		context.Background(),
-		mapFunc,
-	)
-}
```
alfrunes, line 557: I think this deserves a separate commit.

cmd/commands_test.go
```diff
@@ -393,248 +388,3 @@ func TestPropagateStatusesInventory(t *testing.T) {
 		})
 	}
 }
-
-func TestCheckDeviceLimits(t *testing.T) {
```
alfrunes, line 397: ^

store/mongo/datastore_mongo.go
```diff
@@ -114,58 +113,6 @@ func NewDataStoreMongo(config DataStoreMongoConfig) (*DataStoreMongo, error) {
 	return NewDataStoreMongoWithClient(client), nil
 }
 
-func (db *DataStoreMongo) ForEachDatabase(
```
alfrunes, line 118: ^

store/mongo/datastore_mongo.go
```diff
@@ -224,7 +171,7 @@ func (db *DataStoreMongo) GetDevices(
 		findOpts.SetLimit(int64(limit))
 	}
 
-	cursor, err := collDevs.Find(ctx, fltr, findOpts)
+	cursor, err := collDevs.Find(ctx, ctxstore.WithTenantID(ctx, fltr), findOpts)
```
alfrunes, line 174: I think this belongs to the parent commit.

alfrunes (changes requested):
The commit structure is a little off. Also remember we're using semantic commit messages now ;)
I also added some suggestions to the migrations.
There seems to be some issues with the unit- and acceptance tests still.

store/mongo/datastore_mongo.go
```diff
@@ -432,126 +431,6 @@ func (db *DataStoreMongo) DeleteTokenByDevId(
...
-	}
-
-	err = m.Apply(ctx, *ver, migrations)
-	if err != nil {
-		return errors.Wrap(err, "failed to apply migrations")
-	}
-
-	return nil
-}
-
```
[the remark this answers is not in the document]
MaciejTe: Done!

store/mongo/datastore_mongo_test.go
```diff
@@ -847,207 +840,6 @@ func verifyIndexes(t *testing.T, coll *mongo.Collection, expected []mongo.IndexM
...
-				assert.Equal(t, *v, out[len(out)-1].Version)
-
-				// verify that all indexes are created
-			} else {
-				assert.EqualError(t, err, tc.err)
-			}
-		})
-	}
-}
-
```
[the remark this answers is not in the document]
MaciejTe: done!

cmd/commands.go
MaciejTe: Got it!

store/mongo/datastore_mongo.go
MaciejTe: Done!

store/mongo/datastore_mongo.go
MaciejTe: Done!

cmd/commands_test.go
MaciejTe: Done!

store/mongo/migration_2_0_0.go
MaciejTe: Done!

store/mongo/migration_2_0_0.go
MaciejTe: Done!

store/mongo/migration_2_0_0.go
MaciejTe: Thanks! Refactored

store/mongo/migration_2_0_0.go
MaciejTe: done!

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,168 @@
...
+	ds  *DataStoreMongo
+	ctx context.Context
+}
+
+// Up creates an index on status and id in the devices collection
+func (m *migration_2_0_0) Up(from migrate.Version) error {
+	// _false := false
+	// _true := true
+	// var expireAfterSeconds int32 = 0
+	// tokenExpirationIndexName := "TokenExpiration"
```
alfrunes, line 45:
```suggestion
```

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,168 @@
...
+						{Key: model.TenantIDField, Value: 1},
+						{Key: model.DevKeyStatus, Value: 1},
+						{Key: model.DevKeyId, Value: 1},
+					},
+					Options: mopts.Index().
+						SetName(indexDevices_Status).
+						SetUnique(false),
+				}, {
+					Keys: bson.D{
+						{Key: model.DevKeyIdDataSha256, Value: 1},
```
alfrunes, line 66: We need the `tenant_id` top-level key here as well to protect against cross-tenant collisions.

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,168 @@
...
+						{Key: model.TenantIDField, Value: 1},
+						{Key: model.AuthSetKeyIdDataSha256, Value: 1},
+						{Key: model.AuthSetKeyPubKey, Value: 1},
+					},
+					Options: mopts.Index().
+						SetName(indexAuthSet_IdentityDataSha256_PubKey).
+						SetUnique(true),
+				},
+			},
+		},
```
alfrunes, line 87: Reviewing the queries we make to the `auth_sets` collection, we'll also need an index key on the `device_id` field too. That is,
```json
{"tenant_id": 1, "device_id": 1}
```
You may also consider appending the a key on the timestamp as well to get the ability to sort the autsets for a device by date.

store/mongo/migration_2_0_0.go
```diff
@@ -0,0 +1,168 @@
...
+				}
+			}
+			if len(writes) > 0 {
+				_, err := collOut.BulkWrite(m.ctx, writes)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
```
alfrunes, line 162: We also need to handle the tokens collection (but separately). We need to make sure that the TTL index exists for the `exp` token claim. See `migration_1_8_0.go`.

store/mongo/datastore_mongo.go
```diff
@@ -303,12 +300,12 @@ func (db *DataStoreMongo) AddToken(ctx context.Context, t *jwt.Token) error {
...
 
 	filter := bson.M{"_id": t.Claims.ID}
-	update := bson.M{"$set": t}
+	update := bson.M{"$set": ctxstore.WithTenantID(ctx, t)}
 	updateOpts := mopts.Update()
 	updateOpts.SetUpsert(true)
 
 	if _, err := collTokens.UpdateOne(
-		ctx, filter, update, updateOpts,
+		ctx, ctxstore.WithTenantID(ctx, filter), update, updateOpts,
```
alfrunes, line 308: Token claims should not have another tenant_id field. (It's already present in `mender.tenant` field).
```suggestion
		ctx, filter, update, updateOpts,
```

store/mongo/datastore_mongo.go
```diff
@@ -342,7 +338,7 @@ func (db *DataStoreMongo) DeleteToken(ctx context.Context, jti oid.ObjectID) err
 	collTokens := database.Collection(DbTokensColl)
 
 	filter := bson.M{"_id": jti}
-	result, err := collTokens.DeleteOne(ctx, filter)
+	result, err := collTokens.DeleteOne(ctx, ctxstore.WithTenantID(ctx, filter))
```
alfrunes, line 341:
```suggestion
	result, err := collTokens.DeleteOne(ctx, filter)
```

store/mongo/datastore_mongo.go
```diff
@@ -366,7 +362,7 @@ func (db *DataStoreMongo) DeleteTokenByDevId(
 ) error {
 	c := db.client.Database(ctxstore.DbFromContext(ctx, DbName)).
 		Collection(DbTokensColl)
-	ci, err := c.DeleteMany(ctx, bson.M{"sub": devID})
+	ci, err := c.DeleteMany(ctx, ctxstore.WithTenantID(ctx, bson.M{"sub": devID}))
```
alfrunes, line 365:
```suggestion
	var tenantID string
	if id := identity.FromContext(ctx); id != nil {
		tenantID = id.Tenant
	}
	ci, err := c.DeleteMany(ctx, bson.M{"sub": devID, "mender.tenant": tenantID})
```

alfrunes: By the way, there's no index backing this query. We should analyze the usage of this function (NOTE to self).

alfrunes (changes requested):
This is as far as I got today. We need one more index in the `auth_sets` collection and at least keep the TTL index for the tokens collection.

store/mongo/datastore_mongo.go
MaciejTe: `mender.tenant` wouldn't work because of struct `inline` and `omitempty` bson tags. Currently `tenant` field is used to filter tokens.

store/mongo/migration_2_0_0.go
MaciejTe: Done!

store/mongo/datastore_mongo.go
MaciejTe: see above comment

store/mongo/migration_2_0_0.go
MaciejTe: Done!

MaciejTe:

MaciejTe:

kjaskiewiczz (changes requested):
AFAIK device limits will not work - we're using limit name as an _id, so, without introducing some changes here, it will be impossible to set different limits for different tenants

PR #600, opened 2022-06-29
fix: Support PKCS#8 PEM format for JWT Signing key

alfrunes:
Changelog: Title
Ticket: MEN-5673


alfrunes:

keys/key_test.go
```diff
@@ -11,60 +11,153 @@
...
+
+	t.Run("error/file not exist", func(t *testing.T) {
+		path := filepath.Join(t.TempDir(), "not-exist.pem")
+		_, err := LoadRSAPrivate(path)
+		var expected *os.PathError
+		assert.ErrorAs(t, err, &expected)
+	})
 }
+
+const (
```
merlin-northern, line 93: why those consts are defined after the test? 🤔

merlin-northern (approved):
I think we should extend backend integration tests with the PKCS8 keys scenarios for each service to be sure they always work. opinions?

tranchitella (approved):

PR #609, opened 2022-08-19
chore: Align internal Device and AuthSet model with the API model

alfrunes:
Removes the local convertion and reallocation of device data as well as
simplifying the internal interface for using these structs.


tranchitella (approved):

alfrunes:

kjaskiewiczz (approved):

PR #628, opened 2023-02-27
fix(ci): mock `reindex_reporting`

tranchitella:
Changelog: None
Ticket: MEN-6342


kjaskiewiczz (approved):
thanks!
maybe we should think about introducing mmock here (and in other places) in the future

tranchitella:
> thanks! maybe we should think about introducing mmock here (and in other places) in the future

in the future... now, let's fix it.

PR #635, opened 2023-03-25
Dependabot updates

tranchitella:


Dockerfile
```diff
@@ -12,6 +13,7 @@ RUN CGO_ENABLED=0 GOARCH=amd64 go build -o deviceauth .
 FROM scratch
 EXPOSE 8080
 COPY --from=builder /etc_extra/ /etc/
+COPY --from=builder --chown=nobody /tmp_extra/ /tmp/
```
merlin-northern, line 16: why are we coping to /tmp? why do we need tmp_extra?

merlin-northern (changes requested):

Dockerfile
merlin-northern: sorry I see it.

merlin-northern (approved):

PR #648, opened 2023-06-02
fix: do not accept device which is being decommissioned

kjaskiewiczz:
Ticket: MEN-6504


kjaskiewiczz:

merlin-northern (approved):
great find great fix thanks

merlin-northern:
we should cp to 3.4 for sure.

PR #660, opened 2023-07-16
chore(fix/mongo): Migration 2.0.0 overwrites the same limits document

alfrunes:


alfrunes:

store/mongo/migration_2_0_0.go
```diff
@@ -227,28 +227,27 @@ func (m *migration_2_0_0) Up(from migrate.Version) error {
...
-				id := cur.Current.Lookup(dbFieldID)
-				var currentId string
-				err = id.Unmarshal(&currentId)
-				if err != nil {
-					return errors.Wrap(err, "the id found is un-parsable")
+				id, ok := cur.Current.Lookup(dbFieldID).StringValueOK()
+				if !ok {
+					return fmt.Errorf(
+						"found suspicious limits document: %s",
+						cur.Current.String())
```
merlin-northern, line 234: this is just a minor opt, right? boils down to the same thing.

store/mongo/migration_2_0_0.go
```diff
@@ -227,28 +227,27 @@ func (m *migration_2_0_0) Up(from migrate.Version) error {
...
-				item = findAndReplace(item, dbFieldID, oid.NewUUIDv4().String())
-
+				item[dbFieldName] = id
+				item[dbFieldTenantID] = tenantID
+				item[dbFieldID] = primitive.NewObjectID()
 				writes = append(writes, mongo.
 					NewReplaceOneModel().
-					SetFilter(bson.D{{Key: dbFieldName, Value: id}}).
+					SetFilter(bson.D{
+						{Key: dbFieldTenantID, Value: tenantID},
```
merlin-northern, line 246: how can this work? we do not have tenant id in the documents.

alfrunes: Exactly, but we are doing an upsert so if it doesn't match, it will do an insert.

PR #668, opened 2023-09-16
Production pipeline to publish multiplatform build only

oldgiova:
Ticket: QA-613


oldgiova:
@tranchitella if it's ok could you please force-merge it?

tranchitella (approved):

PR #672, opened 2023-10-04
feat: delay JWT token signature and claims validation after cache look-up

tranchitella:
Changelog: feat: optimize the token verification end-point delaying the JWT token signature and claims validation after the cache look-up
Ticket: MEN-5497


tranchitella:

devauth/devauth.go
```diff
@@ -1213,31 +1213,24 @@ func verifyTenantClaim(ctx context.Context, verifyTenant bool, tenant string) er
 }
 
 func (d *DevAuth) VerifyToken(ctx context.Context, raw string) error {
-
 	l := log.FromContext(ctx)
 
 	token := &jwt.Token{}
-
 	err := token.UnmarshalJWT([]byte(raw), d.jwt.FromJWT)
```
kjaskiewiczz, line 1219: I thought this is the place where signature is being validated

tranchitella: I don't understand your comment. The signature verification is implemented in FromJWT, which I'm changing with this PR. It now uses `parser.ParseUnverified`, which doesn't perform the verification. I implemented a new method called `Validate` which performs the signature verification and it is called at a later point in time in the `VerifyToken` function.

kjaskiewiczz: right; thanks for the clarification and sorry for the noise

tranchitella:
Not sure how the integration/enterprise can fail for a PR to `deviceauth`...

mzedel:
@tranchitella the pipeline was started before https://github.com/mendersoftware/integration/pull/2448 was merged, I'm quite sure the failure is because of this 👌

alfrunes (approved):

PR #674, opened 2023-10-09
feat: support for Ed25519 server keys for signing the JWT tokens

tranchitella:
Ticket: MEN-6775
Changelog: Title


tranchitella:

jwt/jwt_ed25519.go
```diff
@@ -0,0 +1,96 @@
...
+type JWTHandlerEd25519 struct {
+	privKey         *ed25519.PrivateKey
+	fallbackPrivKey *ed25519.PrivateKey
+}
+
+func NewJWTHandlerEd25519(privKey *ed25519.PrivateKey,
+	fallbackPrivKey *ed25519.PrivateKey) *JWTHandlerEd25519 {
+	return &JWTHandlerEd25519{
+		privKey:         privKey,
+		fallbackPrivKey: fallbackPrivKey,
```
alfrunes, line 33: Unfortunately this does not allow migrating from RSA to ED25519. I guess we need to move the fallback logic one level up, or make the JWT handler generic.

tranchitella: I reimplemented the fallback using multiple jwtHandler.

tranchitella:

config.yaml
```diff
@@ -72,12 +72,26 @@
 
 # default_tenant_token:  <VALID_TENANT_TOKEN>
 
+# Private key type - used for JWT signing
+# Defaults to: rsa
+# Supported values: rsa, ed25519
+# Overwrite with environment variable: DEVICEAUTH_SERVER_PRIV_KEY_TYPE
+
+# server_priv_key_type: rsa
+
```
alfrunes, line 81: This config option seems a bit redundant and could easily be removed by using the `x509` standard library to parse the private key.

alfrunes (approved):
Looks good. Just one minor remark that I won't consider a blocker.

config.yaml
tranchitella: Something like this?
https://github.com/mendersoftware/deviceauth/pull/674/commits/938be599eaa33eb6eefd5895a86fe7c73ae5fda9

jwt/jwt.go
```diff
@@ -41,76 +51,30 @@ type Handler interface {
...
-	parser := jwtgo.NewParser(jwtgo.WithoutClaimsValidation())
-	jwttoken, _, err := parser.ParseUnverified(tokstr, &Claims{})
-	if err == nil {
-		token := Token{}
-		if claims, ok := jwttoken.Claims.(*Claims); ok {
-			token.Claims = *claims
-			return &token, nil
+	switch block.Type {
+	case pemHeaderPKCS1:
+		privKey, err := keys.LoadRSAPrivate(priv)
```
alfrunes, line 62: Sorry, I missed this line before, but this is what I meant would render the `keys` package no longer needed.
```suggestion
		privKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
```

PR #712, opened 2024-04-25
feat: sync check_in_time with inventory if reporting is disabled

kjaskiewiczz:
Ticket: MEN-7202

### DIFF
--- a/devauth/devauth.go
+++ b/devauth/devauth.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Northern.tech AS
+// Copyright 2024 Northern.tech AS
 //
 //	Licensed under the Apache License, Version 2.0 (the "License");
 //	you may not use this file except in compliance with the License.
@@ -45,6 +45,7 @@ import (
 const (
 	MsgErrDevAuthUnauthorized = "dev auth: unauthorized"
 	MsgErrDevAuthBadRequest   = "dev auth: bad request"
+	InventoryScopeSystem      = "system"
 )
 
 var (
@@ -192,10 +193,9 @@ func (d *DevAuth) setDeviceIdentity(ctx context.Context, dev *model.Device, tena
 			continue
 		}
 		attribute := model.DeviceAttribute{
-			Name:        name,
-			Description: nil,
-			Value:       value,
-			Scope:       "identity",
+			Name:  name,
+			Value: value,
+			Scope: "identity",
 		}
 		attributes[i] = attribute
 		i++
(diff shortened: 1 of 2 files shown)


devauth/devauth.go
```diff
@@ -1754,3 +1762,35 @@ func (d *DevAuth) updateCheckInTime(
...
+	checkInTime *time.Time,
+	deviceId string,
+	tenantId string,
+) error {
+	attributes := []model.DeviceAttribute{
+		{
+			Name:        "check_in_time",
+			Description: nil,
+			Value:       checkInTime,
+			Scope:       "system",
```
merlin-northern, line 1777: two small and perhaps annoying comments: `const InventoryScopeSystem = "system"` maybe? and double check: `system` is the scope we want right?

devauth/devauth.go
```diff
@@ -1754,3 +1762,35 @@ func (d *DevAuth) updateCheckInTime(
...
+func (d *DevAuth) syncCheckInTime(
+	ctx context.Context,
+	checkInTime *time.Time,
+	deviceId string,
+	tenantId string,
+) error {
+	attributes := []model.DeviceAttribute{
+		{
+			Name:        "check_in_time",
+			Description: nil,
```
merlin-northern, line 1775: nit: if we skip this line no one will notice, right?

devauth/devauth.go
```diff
@@ -1741,6 +1741,14 @@ func (d *DevAuth) updateCheckInTime(
...
 				return
 			}
+		} else {
+			// update check-in time in inventory
+			if err := d.syncCheckInTime(ctx, checkInTime, deviceId, tenantId); err != nil {
+				log.FromContext(ctx).Errorf(
+					"failed to synchronize device check-in time with inventory: device %s: %s",
+					deviceId, err.Error(),
+				)
+			}
```
alfrunes, line 1751: I'm not sure if this should be conditional on the reporting service being enabled, because it will eventually break once we get the reporting service.

merlin-northern (approved):
looks good

alfrunes (approved):
Looks ok to me, not sure if the change should depend on the reporting service since this will become a stale timestamp once we enable reporting.

devauth/devauth.go
kjaskiewiczz: The idea here is that the check-in time is provided by device search API served by reporting, but if the reporting is not available, the check-in time is provided by inventory. The UI will use reporting to get check-in time if available and inventory if not.
I don't think we need/want to sync check-in time with both - reporting and inventory.
Does it make sense?

alfrunes: I get the idea, but the data will still be there in inventory search once we enable reporting (disable this feature).

kjaskiewiczz: I don't think it make sense to sync check-in time both with reporting and inventory, once/if we'll introduce reporting.
If we'll decide to not reintroduce reporting, then we can remove the condition.
Once/if we'll be reintroducing reporting with its device search API, we can prepare migration or command which will drop check-in time attributes from the devices in inventory.
What do you think?

kjaskiewiczz:

devauth/devauth.go
kjaskiewiczz: done (const); and yes, system is the right scope

devauth/devauth.go
kjaskiewiczz: done

PR #714, opened 2024-05-02
fix: Preauthorize force behavior applies to existing auth sets

alfrunes:
Updates the behavior of the Preauthorize endpoint if "force" paremeter is set:
 * If an authset already exist, the status will be forced to "preauthorized".
 * If the auth set does not exist, a new one will be created.

Changelog: Commit
Ticket: MEN-7241

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -579,22 +579,61 @@ func (db *DataStoreMongo) MigrateTenant(ctx context.Context, database, version s
 	return nil
 }
 
-func (db *DataStoreMongo) AddAuthSet(ctx context.Context, set model.AuthSet) error {
-	c := db.client.Database(DbName).Collection(DbAuthSetColl)
-
-	if set.Id == "" {
+func fillAuthSet(ctx context.Context, authSet *model.AuthSet) {
+	if authSet.Id == "" {
 		uid := oid.NewUUIDv4()
-		set.Id = uid.String()
+		authSet.Id = uid.String()
 	}
 	id := identity.FromContext(ctx)
 	tenantId := ""
 	if id != nil {
 		tenantId = id.Tenant
 	}
-	set.TenantID = tenantId
+	authSet.TenantID = tenantId
+	if authSet.Timestamp == nil {
+		ts := time.Now()
+		authSet.Timestamp = &ts
+	}
+}
+
+func (db *DataStoreMongo) AddAuthSet(ctx context.Context, set model.AuthSet) error {
+	c := db.client.Database(DbName).Collection(DbAuthSetColl)
+	fillAuthSet(ctx, &set)
 
 	if _, err := c.InsertOne(ctx, set); err != nil {
-		if strings.Contains(err.Error(), "duplicate key error") {
+		if mongo.IsDuplicateKeyError(err) {
+			return store.ErrObjectExists
+		}
+		return errors.Wrap(err, "failed to store device")
+	}
+
+	return nil
+}
+
+// UpsertAuthSetStatus inserts a new auth set and if it exists, ensures the
+// AuthSet.Status matches the one in authSet.
+func (db *DataStoreMongo) UpsertAuthSetStatus(ctx context.Context, authSet *model.AuthSet) error {
+	c := db.client.Database(DbName).Collection(DbAuthSetColl)
+
+	fillAuthSet(ctx, authSet)
+
+	fltr := bson.D{
+		{Key: dbFieldTenantID, Value: authSet.TenantID},
+		{Key: dbFieldIDDataSha, Value: authSet.IdDataSha256},
+		{Key: dbFieldPubKey, Value: authSet.PubKey},
+	}
+	setOnInsert := *authSet
+	setOnInsert.Status = ""
+
+	update := bson.D{
+		{Key: "$setOnInsert", Value: setOnInsert},
+		{Key: "$set", Value: bson.D{
+			{Key: dbFieldStatus, Value: authSet.Status},
+		}},
+	}
+	_, err := c.UpdateOne(ctx, fltr, update, mopts.Update().SetUpsert(true))
+	if err != nil {
+		if mongo.IsDuplicateKeyError(err) {
 			return store.ErrObjectExists
 		}
 		return errors.Wrap(err, "failed to store device")
--- a/store/mongo/datastore_mongo_test.go
+++ b/store/mongo/datastore_mongo_test.go
@@ -1378,6 +1378,59 @@ func TestStoreAuthSet(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestUpsertAuthSet(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping TestGetDevices in short mode.")
+	}
+
+	const (
+		IdData       = `{"foo":"bar"}`
+		IdDataSha256 = "7a38bf81f383f69433ad6e900d35b3e2385593f76a7b7ab5d4355b8ba41ee24b"
+		PubKey       = `-----BEGIN PUBLIC KEY-----
+MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE40NzSJ0FefAHPTDR5eUdEiTIhcztPX58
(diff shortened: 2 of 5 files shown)


alfrunes:

merlin-northern (changes requested):
what we need is simple: https://github.com/mendersoftware/deviceauth-enterprise/pull/279  (already tested, passes the [tests/test_mtls.py](https://github.com/mendersoftware/integration/blob/master/tests/tests/test_mtls.py) pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1278559650 there is no need to change the behaviour in this way.

merlin-northern:
> what we need is simple: [mendersoftware/deviceauth-enterprise#279](https://github.com/mendersoftware/deviceauth-enterprise/pull/279) (already tested, passes the [tests/test_mtls.py](https://github.com/mendersoftware/integration/blob/master/tests/tests/test_mtls.py) pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1278559650 there is no need to change the behaviour in the way proposed here.

I had a chat with Lluis and I was filled in with the situation, treat as suggestion.

store/mongo/datastore_mongo_test.go
```diff
@@ -1378,6 +1378,54 @@ func TestStoreAuthSet(t *testing.T) {
...
+		IdData:       `{"foo":"bar"}`,
+		IdDataSha256: []byte("7a38bf81f383f69433ad6e900d35b3e2385593f76a7b7ab5d4355b8ba41ee24b"),
+		PubKey: `-----BEGIN PUBLIC KEY-----
+MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE40NzSJ0FefAHPTDR5eUdEiTIhcztPX58
+/YPgwRoRydYGw299yP5uH+SEHOZUdIxcTLAvGQME5evgPxNjVmq/NVIeXZunTcYs
+UTnpeomhxO70PoGy64hqCAp4cIGDrVJT
+-----END PUBLIC KEY-----`,
+		Status: "preauthorized",
+	}
+
```
merlin-northern, line 1415: please add an assert that and `newPreauth` and `authSet` are equal up to the status.

store/mongo/datastore_mongo.go
```diff
@@ -603,6 +603,49 @@ func (db *DataStoreMongo) AddAuthSet(ctx context.Context, set model.AuthSet) err
...
+		{Key: "$set", Value: bson.D{
+			{Key: dbFieldStatus, Value: authSet.Status},
+		}},
+	}, mopts.Update().
+		SetUpsert(true)); err != nil {
+		if mongo.IsDuplicateKeyError(err) {
+			return store.ErrObjectExists
+		}
+		return errors.Wrap(err, "failed to store device")
+	}
```
merlin-northern, line 644: please reformat it looks terrible and is unreadable.

store/mongo/datastore_mongo.go
```diff
@@ -603,6 +603,49 @@ func (db *DataStoreMongo) AddAuthSet(ctx context.Context, set model.AuthSet) err
...
+		authSet.Timestamp = &ts
+	}
+
+	fltr := bson.D{
+		{Key: dbFieldTenantID, Value: tenantId},
+		{Key: dbFieldIDDataSha, Value: authSet.IdDataSha256},
+		{Key: dbFieldPubKey, Value: authSet.PubKey},
+	}
+	setOnInsert := *authSet
+	setOnInsert.Status = ""
```
merlin-northern, line 632: why is that empty?

store/mongo/datastore_mongo.go
```diff
@@ -603,6 +603,49 @@ func (db *DataStoreMongo) AddAuthSet(ctx context.Context, set model.AuthSet) err
...
+	if _, err := c.UpdateOne(ctx, fltr, bson.D{
+		{Key: "$setOnInsert", Value: setOnInsert},
+		{Key: "$set", Value: bson.D{
+			{Key: dbFieldStatus, Value: authSet.Status},
+		}},
+	}, mopts.Update().
+		SetUpsert(true)); err != nil {
+		if mongo.IsDuplicateKeyError(err) {
+			return store.ErrObjectExists
```
merlin-northern, line 641: just to double check what if we are preauthorizing an already preauthorized authset?

store/mongo/datastore_mongo.go
```diff
@@ -603,6 +603,49 @@ func (db *DataStoreMongo) AddAuthSet(ctx context.Context, set model.AuthSet) err
...
+	if authSet.Timestamp == nil {
+		ts := time.Now()
+		authSet.Timestamp = &ts
+	}
+
+	fltr := bson.D{
+		{Key: dbFieldTenantID, Value: tenantId},
+		{Key: dbFieldIDDataSha, Value: authSet.IdDataSha256},
+		{Key: dbFieldPubKey, Value: authSet.PubKey},
+	}
```
merlin-northern, line 630: why you have chosen `UpdateOne` over insert with upsert true? in the latter case we could have just added one param to the `AddAuthSet`?

store/mongo/datastore_mongo.go
```diff
@@ -603,6 +603,49 @@ func (db *DataStoreMongo) AddAuthSet(ctx context.Context, set model.AuthSet) err
 	return nil
 }
 
+// UpsertAuthSetStatus inserts a new auth set and if it exists, ensures the
+// AuthSet.Status matches the one in authSet.
+func (db *DataStoreMongo) UpsertAuthSetStatus(ctx context.Context, authSet *model.AuthSet) error {
```
merlin-northern, line 608: general comment: AddAuthSet does a lot of the same, this means that now we have duplicate code for some of that, which can lead to problems in maintaining the code and can be cause for regressions. I would try to create a single place to upsert an authset.

merlin-northern (commented):
looks good in general, left some comments. I think I once clicked on "request changes" and there is no way around it :)

store/mongo/datastore_mongo.go
alfrunes: What do you mean? The code is `go fmt`ed :sweat_smile:


I agree, that looks like garbage :hankey:

store/mongo/datastore_mongo.go
alfrunes: I have to clear the status on the `$setOnInsert` since it is set in the `$set` clause. Putting a value in both will lead to a ConflictingValue exception (even if the values are the same). The resulting document that ends up in the DB is always the merged object.

store/mongo/datastore_mongo.go
alfrunes: It will behave the same as doing an update with `{$set: {status: "preauthorized"}}` so the document will not be modified.

store/mongo/datastore_mongo.go
alfrunes: I'm not sure I understand. Did you mean `ReplaceOne` with upsert?
Doing a full replace will update the (inserted) `ts` attribute and potentially other immutable data.

store/mongo/datastore_mongo_test.go
alfrunes: I believe this is the assertion made here: https://github.com/mendersoftware/deviceauth/blob/8f12c84773a3aba9d9b4b4878b1b967cf573d327/store/mongo/datastore_mongo_test.go#L1425
I'll tidy up the test a little and add some comments to make it easier to read.

store/mongo/datastore_mongo.go
alfrunes: I put the shared code into a new function for populating the missing fields in the AuthSet.

merlin-northern (approved):
thanks for listening to my comments :> looks good (I would add units for `fillAuthSet` but lets say I did not notice the lack of them)

alfrunes:
> thanks for listening to my comments :> looks good (I would add units for `fillAuthSet` but lets say I did not notice the lack of them)

:see_no_evil:

store/mongo/datastore_mongo.go
merlin-northern: haha :)

PR #719, opened 2024-06-12
fix: Inconsistent device check_in_time when listing devices

alfrunes:
The lookup for `check_in_time` from the cache does not work when running
Redis in cluster mode because of the MGET command requires keys to hash
to the same slot. This commit replaces MGET with multiple batched GET
commands when running Redis in cluster mode.

Changelog: Commit
Ticket: MEN-7337


alfrunes:

kjaskiewiczz (approved):
looks good

alfrunes:
I also tested it locally on a Redis cluster and it's now working as expected after the last force-push.

